PR #59, opened 2026-06-12
feat: add gRPC warmup support with server reflection

### LINKED ISSUE #25: Add gRPC warmup support
## Context

Modern microservice architectures increasingly use gRPC for inter-service communication. To warm up gRPC services, kube-booster needs to send gRPC requests in addition to HTTP. This phase introduces a protocol-agnostic `Sender` interface with HTTP and gRPC implementations, using server reflection for dynamic method invocation.

This is **Phase 3** of the warmup engine replacement.
- Phase 1: #23 — Replace Vegeta with custom HTTP executor
- Phase 2: #24 — Cross-pod concurrency control
- Phase 4: #26 — Scenario-based execution with CRD

**Depends on:** #24

## New Annotations

| Annotation | Values | Default |
|---|---|---|
| `kube-booster.io/warmup-protocol` | `"http"`, `"grpc"` | `"http"` |
| `kube-booster.io/warmup-grpc-method` | `"package.Service/Method"` | required for gRPC |
| `kube-booster.io/warmup-grpc-payload` | JSON string | `"{}"` |

### Example Usage

```yaml
annotations:
  kube-booster.io/warmup: "enabled"
  kube-booster.io/warmup-protocol: "grpc"
  kube-booster.io/warmup-port: "50051"
  kube-booster.io/warmup-grpc-method: "myapp.CacheService/WarmCache"
  kube-booster.io/warmup-grpc-payload: '{"region": "us-east-1"}'
  kube-booster.io/warmup-requests: "5"
  kube-booster.io/warmup-duration: "30s"
```

### Sender Interface (`pkg/warmup/sender.go`)

```go
type Sender interface {
    Send(ctx context.Context, target Target) (*Response, error)
    Close() error
}

type Target struct {
    URL     string            // full URL (HTTP) or host:port (gRPC)
    Method  string            // HTTP method or "package.Service/Method"
    Headers map[string]string
    Payload []byte
}

type Response struct {
    StatusCode int
    Duration   time.Duration
    Body       []byte  // for response chaining in Phase 4
    Error      error
}
```

### HTTPSender (`pkg/warmup/http_sender.go`)

Extracted from current HTTPExecutor's inline HTTP client logic.

### GRPCSender (`pkg/warmup/grpc_sender.go`)

Uses server reflection for dynamic method invocation:
1. Connect to `podIP:port` with insecure credentials
2. Use `grpcreflect.Client` to discover method descriptor
3. Build `dynamic.Message` from JSON payload
4. Invoke via `grpcdynamic.Stub` or `conn.Invoke()`
5. Return response body as JSON

**Requirements on target service:**
- Must have gRPC server reflection enabled
- Only unary RPCs supported (streaming is out of scope)

### Executor Refactoring

The executor (renamed `HTTPExecutor` → `WarmupExecutor`) retains its ASAP back-to-back loop. The change is that each iteration calls `sender.Send()` instead of `http.Client.Do()`. The sender is created based on `config.Protocol` at the start of `Execute()`.

## Files to Change

| Action | File | What |
|--------|------|------|
| Create | `pkg/warmup/sender.go` | `Sender` interface, `Target`, `Response` types |
| Create | `pkg/warmup/http_sender.go` | `HTTPSender` using `net/http` |
| Create | `pkg/warmup/http_sender_test.go` | Tests |
| Create | `pkg/warmup/grpc_sender.go` | `GRPCSender` using `grpc` + `protoreflect` |
| Create | `pkg/warmup/grpc_sender_test.go` | Tests using `grpc/test/bufconn` |
| Rename+Modify | `pkg/warmup/http_executor.go` → `warmup_executor.go` | Refactor to use Sender; dispatch based on protocol |
| Rename+Modify | `pkg/warmup/http_executor_test.go` → `warmup_executor_test.go` | Add gRPC test cases |
| Modify | `pkg/warmup/config.go` | Add `Protocol`, `GRPCMethod`, `GRPCPayload` fields and parsing |
| Modify | `pkg/warmup/config_test.go` | Add protocol parsing tests |
| Modify | `pkg/webhook/constants.go` | Add 3 new annotation constants |
| Modify | `cmd/controller/main.go` | `NewHTTPExecutor` → `NewWarmupExecutor` |
| Modify | `go.mod` | Add `google.golang.org/grpc`, `github.com/jhump/protoreflect` |

## Config Additions (`pkg/warmup/config.go`)

```go
const (
    ProtocolHTTP = "http"
    ProtocolGRPC = "grpc"
)

type Config struct {
    // ...existing fields...
    Protocol    string // "http" (default) or "grpc"
    GRPCMethod  string // "package.Service/Method"
    GRPCPayload string // JSON string
}

func (c *Config) BuildGRPCAddress() string  // returns "podIP:port"
```

Validation: `GRPCMethod` required when `Protocol == "grpc"`, error otherwise.

## Edge Cases

- gRPC server without reflection → sender creation fails; fail-open
- Invalid gRPC method format → error response; fail-open
- gRPC connection timeout → respects context deadline
- Protocol annotation missing → defaults to HTTP (backward compatible)
- Invalid protocol value → config parse error; fail-open

## Tests

- HTTP sender: httptest.Server, verify headers/status
- gRPC sender: bufconn in-memory server with reflection, test unary call with JSON payload
- Config: protocol annotation parsing, required GRPCMethod validation, invalid protocol
- Executor: protocol dispatch (HTTP path unchanged, gRPC path with test server)

## Verification

```bash
make test && make lint && make build
# Verify HTTP-only annotations still work (backward compatible)
```

## Documentation Updates

- `docs/USAGE.md`: Add gRPC warmup section with annotation examples
- `docs/OBSERVABILITY.md`: Note that `warmup_requests_total` includes gRPC requests
- `config/samples/sample_grpc_deployment.yaml`: Example gRPC deployment with warmup

hhiroshell:
## Summary

Implements #25 — gRPC warmup support as Phase 3 of the warmup engine.

- Introduces a protocol-agnostic `Sender` interface (`pkg/warmup/sender.go`) with `HTTPSender` and `GRPCSender` implementations
- `GRPCSender` uses gRPC server reflection (`grpc_reflection_v1`) and `dynamicpb` to invoke any unary method at runtime — no compiled proto files required
- Refactors `HTTPExecutor` → `WarmupExecutor`, which dispatches to the correct sender based on `Config.Protocol`
- Three new pod annotations: `warmup-protocol`, `warmup-grpc-method`, `warmup-grpc-payload`
- Fully backward-compatible: pods without `warmup-protocol` continue to use HTTP

## Changes Made

- `pkg/webhook/constants.go` — 3 new annotation constants
- `pkg/warmup/config.go` — `Protocol`, `GRPCMethod`, `GRPCPayload` fields + parsing + validation + `BuildGRPCAddress()`
- `pkg/warmup/sender.go` — new `Sender` interface, `Target`, `Response` types
- `pkg/warmup/http_sender.go` — HTTP logic extracted from executor
- `pkg/warmup/grpc_sender.go` — gRPC reflection client + dynamicpb invocation (pure `google.golang.org/grpc` + `google.golang.org/protobuf`, no new dependencies)
- `pkg/warmup/warmup_executor.go` — renamed from `http_executor.go`; protocol-dispatching loop
- `cmd/controller/main.go` — `NewHTTPExecutor` → `NewWarmupExecutor`
- `config/samples/sample_grpc_deployment.yaml` — example gRPC workload
- Tests: `http_sender_test.go`, `grpc_sender_test.go`, `warmup_executor_test.go`, `config_test.go`

## Testing

- `make test`: all packages pass, warmup package at 82.6% coverage
- `make lint`: 0 issues
- `make build`: clean build
- `TestGRPCSender_Send_Success`: happy path against real in-process gRPC server with reflection
- `TestGRPCSender_Send_ReflectionUnavailable`: fail-open path (no reflection → error response, no panic)
- `TestWarmupExecutor_Execute_GRPC`: end-to-end gRPC warmup through the executor
- `TestWarmupExecutor_Execute_GRPC_NoReflection`: executor fail-open with unavailable reflection

## Related Issue

Closes #25

---
*PR generated with [Claude Code](https://claude.ai/code) - github-devflow:implement*


pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+//   - Application-level gRPC error (non-OK status) → StatusCode 500, Error nil
+//     (latency is recorded; the application processed the request)
+//   - Transport/reflection failure → Error non-nil (latency is not recorded)
+func (s *GRPCSender) Send(ctx context.Context, target Target) *Response {
+	start := time.Now()
+
+	// Lazy dial on first Send.
+	if s.conn == nil {
+		conn, err := grpc.NewClient(target.Address,
+			grpc.WithTransportCredentials(insecure.NewCredentials()))
```
hhiroshell, line 59: **[Security]** :warning: gRPC connections use `insecure.NewCredentials()`, so all warmup traffic between controller and pod is plaintext with no transport auth (CWE-319). Pod-to-pod is often trusted, but this hard-codes no-TLS with no config knob. Consider an optional TLS config or documenting the threat model.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+	"google.golang.org/protobuf/types/dynamicpb"
+)
+
+// GRPCSender executes a single gRPC unary warmup request using server reflection.
+// It dials lazily on the first Send call and reuses the connection and cached method
+// descriptor for subsequent calls.
+type GRPCSender struct {
+	logger     logr.Logger
+	conn       *grpc.ClientConn
+	methodDesc protoreflect.MethodDescriptor // cached after first successful reflection lookup
```
hhiroshell: **[Design]** :information_source: `conn`/`methodDesc` are cached as struct fields, but a fresh `GRPCSender` is created on every `Execute()` (warmup_executor.go:68). The cache only helps within a single pod's warmup loop — there is no cross-pod connection reuse. Either promote caching to the executor level or document `GRPCSender` as a single-use, per-Execute object so the cache scope isn't misleading.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+		conn, err := grpc.NewClient(target.Address,
+			grpc.WithTransportCredentials(insecure.NewCredentials()))
+		if err != nil {
+			return &Response{Error: fmt.Errorf("grpc dial %s: %w", target.Address, err), Duration: time.Since(start)}
+		}
+		s.conn = conn
+	}
+
+	// Lazy reflection lookup: discover and cache the method descriptor.
+	if s.methodDesc == nil {
```
hhiroshell: **[Logic]** :warning: When `resolveMethodDescriptor` fails (e.g. reflection unavailable), `s.conn` is set but `s.methodDesc` stays nil, so every subsequent `Send` in the request loop re-enters this branch and retries the reflection lookup. A permanently-failing reflection endpoint causes one full round-trip per request before the loop ends. Set a `reflectionFailed` sentinel after the first failure and return immediately on later calls.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+		md, err := resolveMethodDescriptor(ctx, s.conn, serviceSymbol, methodName)
+		if err != nil {
+			return &Response{Error: err, Duration: time.Since(start)}
+		}
+		s.methodDesc = md
+	}
+
+	// Build request message from JSON payload.
+	reqMsg := dynamicpb.NewMessage(s.methodDesc.Input())
```
hhiroshell: **[Performance]** :rotating_light: `dynamicpb.NewMessage(s.methodDesc.Input())` allocates a new request message on every `Send()`. With a large RequestCount this is thousands of allocations and GC pressure. Allocate once in `NewGRPCSender` and call `msg.Reset()` before each `Unmarshal`.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+	if len(target.Payload) > 0 {
+		if err := protojson.Unmarshal(target.Payload, reqMsg); err != nil {
+			return &Response{Error: fmt.Errorf("invalid gRPC payload: %w", err), Duration: time.Since(start)}
+		}
+	}
+
+	// Invoke unary RPC.
+	respMsg := dynamicpb.NewMessage(s.methodDesc.Output())
```
hhiroshell: **[Performance]** :rotating_light: `dynamicpb.NewMessage(s.methodDesc.Output())` allocates a fresh response message per `Send()`. Create once and `Reset()` per request to cut allocations proportional to RequestCount.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+	methodPath := "/" + string(s.methodDesc.Parent().FullName()) + "/" + string(s.methodDesc.Name())
```
hhiroshell: **[Performance]** :information_source: `methodPath` is rebuilt by string concatenation on every `Send()`. Since `methodDesc` is cached after the first reflection lookup, compute `methodPath` once and cache it on the struct.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+	if s.conn != nil {
+		return s.conn.Close()
+	}
+	return nil
+}
+
+// parseGRPCMethod splits "package.Service/Method" into ("package.Service", "Method").
+func parseGRPCMethod(fullMethod string) (serviceSymbol, methodName string, err error) {
+	idx := strings.LastIndex(fullMethod, "/")
+	if idx < 0 || idx == len(fullMethod)-1 {
```
hhiroshell: **[Logic]** :warning: `parseGRPCMethod` does not reject a leading-slash method like `"/Check"`. When `idx == 0` the guard `idx < 0 || idx == len(fullMethod)-1` is false, so it returns an empty `serviceSymbol` with no error, which then produces a cryptic reflection error downstream. Add `|| idx == 0` to reject a leading slash early.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+	if md.IsStreamingClient() || md.IsStreamingServer() {
+		return nil, fmt.Errorf("only unary RPCs are supported; %q is a streaming method", serviceSymbol+"/"+methodName)
+	}
+
+	return md, nil
+}
+
+// fetchFileDescriptorBytes uses the gRPC server reflection service to retrieve the
+// serialized FileDescriptorProto bytes for the file containing serviceSymbol.
+func fetchFileDescriptorBytes(ctx context.Context, conn *grpc.ClientConn, serviceSymbol string) ([][]byte, error) {
```
hhiroshell, line 174: **[Security]** :warning: `fetchFileDescriptorBytes`/`buildFileSet` process serialized `FileDescriptorProto` bytes from the target pod's reflection endpoint with no cap on size or descriptor count (CWE-400). A compromised application container could return an arbitrarily large reflection response and exhaust controller memory, affecting warmup cluster-wide. Add a hard size limit before unmarshaling.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+		}
+		resolver := &fileSetResolver{local: files}
+		fd, err := protodesc.NewFile(fdp, resolver)
+		if err != nil {
+			return fmt.Errorf("failed to build descriptor for %q: %w", name, err)
+		}
+		if err := files.RegisterFile(fd); err != nil {
+			// Duplicate registration is harmless (can occur with shared transitive deps).
+			if !strings.Contains(err.Error(), "already registered") {
```
hhiroshell: **[Security]** :information_source: **[Design]** :information_source: The duplicate-registration guard uses `strings.Contains(err.Error(), "already registered")`, matching an internal, non-stable error string from `google.golang.org/protobuf`. If upstream changes the wording, registration errors would be silently swallowed. Prefer `errors.As` with a sentinel, or check the `registered` map / `FindFileByPath` before `RegisterFile` to avoid the duplicate path entirely.

pkg/warmup/warmup_executor.go
```diff
@@ -15,28 +14,29 @@ type Executor interface {
...
 	}
 }
 
-// HTTPExecutor fires warmup requests back-to-back (ASAP model) using net/http.
-type HTTPExecutor struct {
+// WarmupExecutor fires warmup requests back-to-back (ASAP model), dispatching to the
+// appropriate Sender based on the configured protocol (HTTP or gRPC).
+type WarmupExecutor struct {
 	logger      logr.Logger
 	client      *http.Client
```
hhiroshell, line 34: **[Design]** :warning: `WarmupExecutor` owns an `*http.Client` field that the gRPC path never uses (it builds its own `GRPCSender`). A protocol-agnostic type carrying HTTP-only plumbing means dead state as protocols grow. Consider depending on a `Sender` factory or a pre-built `Sender` instead of owning the `http.Client` directly.

pkg/warmup/warmup_executor.go
```diff
@@ -48,28 +48,57 @@ func NewHTTPExecutor(logger logr.Logger, opts ...HTTPExecutorOption) *HTTPExecut
...
-		"requestCount", config.RequestCount,
-		"timeout", config.Timeout)
+	// Choose sender and build target based on protocol.
+	var (
+		sender Sender
+		target Target
+	)
+	switch config.Protocol {
+	case ProtocolGRPC:
+		sender = NewGRPCSender(e.logger)
```
hhiroshell, line 69: **[Design]** :warning: **[Performance]** :warning: A new `GRPCSender` — and thus a new gRPC connection — is created inside `Execute()` for every pod warmup. The sender's lazy-dial/reflection caching never benefits across executions, and at scale this means many independent connections to the same target. There's also no way to inject a pre-built/mock `Sender`, which makes `WarmupExecutor` hard to unit-test. Accept a `Sender` factory or pre-built sender (enables pooling-by-target and testability).

pkg/warmup/warmup_executor.go
```diff
@@ -48,28 +48,57 @@ func NewHTTPExecutor(logger logr.Logger, opts ...HTTPExecutorOption) *HTTPExecut
...
+			Payload: []byte(config.GRPCPayload),
+		}
+		e.logger.V(1).Info("starting gRPC warmup",
+			"pod", config.PodName,
+			"namespace", config.PodNamespace,
+			"address", target.Address,
+			"method", target.Method,
+			"requestCount", config.RequestCount,
+			"timeout", config.Timeout)
+	default:
```
hhiroshell: **[Design]** :warning: The protocol `switch` uses `default` for the HTTP case rather than an explicit `case ProtocolHTTP`. If a third protocol is added and a case is forgotten, it silently falls back to HTTP instead of erroring. Use `case ProtocolHTTP:` and move the unknown-protocol error to `default`, mirroring the validation in `ParseConfig`.

pkg/warmup/warmup_executor.go
```diff
@@ -85,43 +114,23 @@ func (e *HTTPExecutor) Execute(ctx context.Context, config *Config) *Result {
 
 		if err := e.rateLimiter.Wait(warmupCtx); err != nil {
 			// Context cancelled or deadline exceeded before token was available.
```
hhiroshell: **[Git History]** :information_source: This comment was shortened during the rename from `http_executor.go`. The original (commit 0ca30b3) also explained that remaining un-attempted requests are counted as failed so the result message and fail-open decision reflect the full picture. Logic is unchanged, but consider restoring the fuller comment.

hhiroshell (commented):

This review was generated by analyzing PR #59 from 8 perspectives. 21 findings are posted as inline line comments; documentation/convention findings (on files not in this diff) are summarized below.

### Summary

| Perspective | Findings |
|-------------|----------|
| Logic & Correctness | 5 issues |
| Design & Maintainability | 8 issues |
| Security | 4 issues |
| Performance | 4 issues |
| Convention Compliance | 3 issues |
| Git History Context | 2 issues |
| PR History Context | 8 notes (mostly positive) |
| Documentation | 16 issues |
| **Total** | **50** |

Inline severity breakdown: 🚨 2 error · ⚠️ 10 warning · ℹ️ 9 info.

### Key Findings

**🚨 Performance — per-request message allocation (grpc_sender.go:74, :82)**
`dynamicpb.NewMessage(...)` allocates a fresh request and response message on every `Send()`. Across a large RequestCount this is the most impactful issue in the PR: allocate once and `Reset()` per request.

**⚠️ Correctness — gRPC method/payload validated too late (config.go:135, grpc_sender.go:117)**
`GRPCMethod` is only emptiness-checked at parse time; malformed values like `"/Check"` or `"NoSlash"` pass `ParseConfig` and fail with a cryptic error at warmup time. `parseGRPCMethod` also fails to reject a leading slash (`idx == 0`). Validate format in `ParseConfig`.

**⚠️ Correctness — reflection failure retried every request (grpc_sender.go:61, warmup_executor.go:123)**
A permanently-failing reflection endpoint is retried once per request because `methodDesc` stays nil, and the executor loop never breaks on persistent transport errors — the full Timeout can be consumed on a non-recoverable condition.

**⚠️ Design — sender/connection created per Execute (warmup_executor.go:33, :68, :81)**
`WarmupExecutor` builds a new `GRPCSender` (and gRPC connection) on every pod warmup, so its caching has no cross-pod benefit and there's no way to inject a mock `Sender`. The protocol `switch` also routes HTTP through `default` rather than an explicit case.

**⚠️ Security — no TLS, unbounded reflection response (grpc_sender.go:53, :159)**
gRPC traffic uses `insecure.NewCredentials()` (plaintext, no config knob), and reflection responses from the target pod are processed with no size cap — a compromised container could exhaust controller memory (cluster-wide impact).

**⚠️ Git History — breaking API rename (main.go:133)**
`NewHTTPExecutor` → `NewWarmupExecutor`. In-repo call sites updated and tests pass; relevant only to external consumers.

### Documentation (16 issues — files outside this diff, cannot be inline)

The new annotations (`warmup-protocol`, `warmup-grpc-method`, `warmup-grpc-payload`) and gRPC support are not reflected in user/dev docs. Per CLAUDE.md ("New annotations/config → Update all docs"), update:

- **docs/USAGE.md** — add the 3 gRPC annotations to the Configuration Annotations table; add a gRPC warmup example; clarify HTTP-only notes (custom headers, "How It Works" flow).
- **docs/IMPLEMENTATION_SUMMARY.md** — gRPC is now listed under "What's NOT Implemented" / "Next Steps" / "Known Limitations (HTTP only)"; these are now incorrect and must be updated.
- **docs/DEVELOPMENT.md** — "Future Development → gRPC Support" should move to implemented.
- **CLAUDE.md** — annotation example and Implementation Pattern step 6 ("sends HTTP warmup requests") should mention both protocols. *(Convention reviewer flagged the same docs gaps as errors.)*
- **README.md** — Features list and config example still HTTP-only.

### PR History Context (positive)

The refactor aligns with guidance from PR #27: the `Sender` interface and `Response` type now live in their own `sender.go` (the stable contract), concrete senders are split into `http_sender.go`/`grpc_sender.go`, and the shared `http.Client` connection pooling is preserved across executions. PR #27's bounds on `warmup-requests`/`warmup-timeout` remain intact; the open follow-up is bounding the gRPC payload/method (covered by the inline findings above).

---

pkg/warmup/warmup_executor.go
```diff
@@ -85,43 +114,23 @@ func (e *HTTPExecutor) Execute(ctx context.Context, config *Config) *Result {
...
-		}
-		req.Header.Set("User-Agent", "kube-booster/1.0")
-		req.Header.Set("X-Warmup-Request", "true")
+		resp := sender.Send(warmupCtx, target)
 
-		resp, err := e.client.Do(req)
-		latency := time.Since(reqStart)
-
-		if err != nil {
+		if resp.Error != nil {
```
hhiroshell, line 123: **[Logic]** :warning: On a persistent non-context error from `sender.Send` (e.g. reflection permanently unavailable, pod IP unreachable), the loop does not break and retries for all RequestCount iterations. For gRPC each failed attempt is a full round-trip/timeout, so the entire Timeout can be burned on a clearly non-recoverable condition. Not a regression from HTTP, but consider breaking early on clearly-permanent transport errors.

pkg/warmup/config.go
```diff
@@ -49,6 +58,15 @@ type Config struct {
 
 	// Port is the port to use for warmup requests
 	Port int
+
+	// Protocol is the warmup protocol: "http" (default) or "grpc"
```
hhiroshell, line 63: **[Design]** :information_source: `Config` mixes HTTP-specific (`Endpoint`) and gRPC-specific (`GRPCMethod`, `GRPCPayload`) fields in a flat, sparsely-populated struct; each added protocol modifies the same type/`ParseConfig`. Acceptable for an early-stage API given the good field docs, but embedding protocol sub-structs (`HTTPConfig`/`GRPCConfig`) would scale better.

pkg/warmup/config.go
```diff
@@ -101,6 +121,32 @@ func ParseConfig(pod *corev1.Pod) (*Config, error) {
...
+			switch protocol {
+			case ProtocolHTTP, ProtocolGRPC:
+				config.Protocol = protocol
+			default:
+				return config, fmt.Errorf("invalid warmup-protocol value %q: must be %q or %q", protocol, ProtocolHTTP, ProtocolGRPC)
+			}
+		}
+
+		// Parse gRPC method
+		if method, ok := annotations[webhook.AnnotationWarmupGRPCMethod]; ok && method != "" {
```
hhiroshell, line 136: **[Logic]** :warning: The `GRPCMethod` annotation is stored verbatim and only checked for emptiness (line 145); a malformed value like `"/Check"` or `"NoSlash"` passes `ParseConfig` and only fails at runtime in `Send`. Validate the format here by calling `parseGRPCMethod` and returning a descriptive error immediately.

pkg/warmup/config.go
```diff
@@ -101,6 +121,32 @@ func ParseConfig(pod *corev1.Pod) (*Config, error) {
...
+			config.GRPCMethod = method
+		}
+
+		// Parse gRPC payload
+		if payload, ok := annotations[webhook.AnnotationWarmupGRPCPayload]; ok && payload != "" {
```
hhiroshell, line 144: **[Security]** :information_source: The `warmup-grpc-payload` annotation is stored without JSON validation and only parsed by `protojson.Unmarshal` at warmup time. Fail-open means it won't block readiness, but validating well-formedness in `ParseConfig` surfaces misconfiguration earlier and matches how other annotations are validated (CWE-20).

pkg/warmup/sender.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Target describes a single warmup request.
+type Target struct {
+	// Address is "host:port" for gRPC or a full URL for HTTP.
+	Address string
+
+	// Method is the HTTP verb (e.g. "GET") or gRPC method ("package.Service/Method").
+	Method string
+
+	// Headers contains additional request headers (HTTP only).
```
hhiroshell: **[Design]** :information_source: `Target` is shared across senders but field meaning shifts by protocol: `Headers`/`Method`(verb) are HTTP-only while `Payload`/`Method`(method path) drive gRPC. The doc comments help; consider stating explicitly which fields are ignored by each sender (e.g. "ignored by GRPCSender").

pkg/warmup/grpc_sender_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+func TestGRPCSender_Close_NoConn(t *testing.T) {
+	sender := NewGRPCSender(ctrl.Log.WithName("test"))
+	if err := sender.Close(); err != nil {
+		t.Errorf("Close() on unconnected sender returned error: %v", err)
+	}
+}
+
+// Ensure the test server address helper works.
+var _ = fmt.Sprintf // suppress unused import warning
```
hhiroshell: **[Design]** :information_source: `var _ = fmt.Sprintf // suppress unused import warning` silences an unused `fmt` import rather than removing it. Drop the import (and this line) instead — the suppression is misleading to future contributors.

pkg/warmup/config_test.go
```diff
@@ -367,6 +465,15 @@ func TestParseConfig(t *testing.T) {
 			if config.Port != tt.wantConfig.Port {
 				t.Errorf("Port = %v, want %v", config.Port, tt.wantConfig.Port)
 			}
+			if tt.wantConfig.Protocol != "" && config.Protocol != tt.wantConfig.Protocol {
```
hhiroshell: **[Logic]** :information_source: The Protocol/GRPCMethod/GRPCPayload assertions are guarded by `tt.wantConfig.X != ""`, so pre-existing cases with zero-value `wantConfig` never verify the new default `Protocol == ProtocolHTTP`, and a case couldn't assert "protocol must be empty". Make the equality checks unconditional, or always populate `wantConfig.Protocol`.

cmd/controller/main.go
```diff
@@ -130,7 +130,7 @@ func main() {
 	rateLimiter := warmup.NewRequestRateLimiter(float64(maxWarmupRPS))
 
 	// Create warmup executor
-	warmupExecutor := warmup.NewHTTPExecutor(ctrl.Log.WithName("warmup"),
+	warmupExecutor := warmup.NewWarmupExecutor(ctrl.Log.WithName("warmup"),
```
hhiroshell, line 133: **[Git History]** :warning: `NewHTTPExecutor` → `NewWarmupExecutor` is a breaking public-API rename (intentional, part of the protocol-agnostic refactor). All in-repo call sites are updated and tests pass; flagging for any external consumers of the package.

hhiroshell: At this stage of the project, changing public APIs is not an issue.

config/samples/sample_grpc_deployment.yaml
```diff
@@ -0,0 +1,43 @@
...
+      labels:
+        app: my-grpc-app
+      annotations:
+        # Enable warmup using gRPC protocol
+        kube-booster.io/warmup: "enabled"
+        kube-booster.io/warmup-protocol: "grpc"
+
+        # Fully-qualified gRPC method to invoke during warmup.
+        # The server must expose this method via gRPC server reflection.
+        # Common choices: a health-check or a lightweight read-only RPC.
```
hhiroshell: For example, a health-check endpoint is fine. However, I believe that in most use cases, users will use practical endpoints for wamup.

pkg/warmup/config.go
hhiroshell: Make an issue for this.

pkg/warmup/config.go
hhiroshell: What do you think about performing the validation in the kube-booster webhook? It will let users know about errors in the annotations at an earlier stage.

pkg/warmup/warmup_executor.go
hhiroshell: Create a refactoring issue based on this comment.

pkg/warmup/warmup_executor.go
hhiroshell: Good catch. We should consider addressing this comment in this PR.

pkg/warmup/warmup_executor.go
hhiroshell: What do you think about creating new error structs that allow WarmupExecutor to decide whether it should break the loop or not?

pkg/warmup/warmup_executor.go
```diff
@@ -48,28 +48,57 @@ func NewHTTPExecutor(logger logr.Logger, opts ...HTTPExecutorOption) *HTTPExecut
...
+				"User-Agent":       "kube-booster/1.0",
+				"X-Warmup-Request": "true",
+			},
+		}
+		e.logger.V(1).Info("starting HTTP warmup",
+			"pod", config.PodName,
+			"namespace", config.PodNamespace,
+			"endpoint", endpoint,
+			"requestCount", config.RequestCount,
+			"timeout", config.Timeout)
```
hhiroshell: We don't need to write separate logging lines for each protocol. The protocol type can be obtained from the config.

pkg/warmup/http_sender_test.go
```diff
@@ -0,0 +1,106 @@
...
+		wantErr        bool
+	}{
+		{
+			name: "success 200",
+			serverHandler: func(w http.ResponseWriter, r *http.Request) {
+				if r.Header.Get("X-Warmup-Request") != "true" {
+					t.Errorf("expected X-Warmup-Request: true header")
+				}
+				if r.Header.Get("User-Agent") != "kube-booster/1.0" {
+					t.Errorf("expected User-Agent: kube-booster/1.0 header")
```
hhiroshell, line 29: The `/1.0` part of this header should be automatically updated based on the kube-booster version.

pkg/warmup/grpc_sender.go
hhiroshell: Additionally, please make an issue that addresses this comment.

pkg/warmup/grpc_sender.go
hhiroshell: Please create an issue that addresses this comment.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+			return &Response{Error: fmt.Errorf("grpc dial %s: %w", target.Address, err), Duration: time.Since(start)}
+		}
+		s.conn = conn
+	}
+
+	// Lazy reflection lookup: discover and cache the method descriptor.
+	if s.methodDesc == nil {
+		serviceSymbol, methodName, err := parseGRPCMethod(target.Method)
+		if err != nil {
+			return &Response{Error: err, Duration: time.Since(start)}
```
hhiroshell, line 76: This Duration doesn't reflect the communication between the pod and kube-booster. Is it meaningful enough to be included in the response?

pkg/warmup/warmup_executor.go
hhiroshell: I realized that it decides whether to stop the loop or not based on whether an error is received.

pkg/warmup/grpc_sender.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil, fmt.Errorf("symbol %q is not a service", serviceSymbol)
+	}
+
+	md := svcDesc.Methods().ByName(protoreflect.Name(methodName))
+	if md == nil {
+		return nil, fmt.Errorf("method %q not found in service %q", methodName, serviceSymbol)
+	}
+
+	if md.IsStreamingClient() || md.IsStreamingServer() {
+		return nil, fmt.Errorf("only unary RPCs are supported; %q is a streaming method", serviceSymbol+"/"+methodName)
```
hhiroshell, line 166: We should write this constraint somewhere in the docs.

pkg/warmup/grpc_sender_test.go
```diff
@@ -0,0 +1,233 @@
...
+	// Send multiple requests; connection and method descriptor should be reused.
+	for i := 0; i < 3; i++ {
+		resp := sender.Send(context.Background(), target)
+		if resp.Error != nil {
+			t.Fatalf("Send() request %d unexpected error: %v", i+1, resp.Error)
+		}
+		if resp.StatusCode != 200 {
+			t.Errorf("Send() request %d StatusCode = %d, want 200", i+1, resp.StatusCode)
+		}
+	}
```
hhiroshell: Can it check the connection reuse?

pkg/warmup/grpc_sender.go
hhiroshell: Please also document this constraint.

hhiroshell: Acknowledged — this is a real gap. The `insecure.NewCredentials()` path is hard-coded with no config knob, so there's no way for a cluster operator to enable mTLS even if the cluster's network policy doesn't provide pod-level isolation.

I'll create a follow-up issue to add an optional TLS configuration (likely a flag like `--grpc-tls-cert` / `--grpc-tls-key` with a secure-by-default opt-in path). In the meantime I'll document the plaintext-only constraint in `docs/USAGE.md` under the gRPC warmup section, covering the threat model and the recommendation to apply a NetworkPolicy that restricts controller-to-pod traffic.

---

pkg/warmup/grpc_sender.go
hhiroshell: Agreed — the caching comment on the struct fields is misleading because `GRPCSender` is instantiated fresh inside `Execute()` on every pod warmup (warmup_executor.go:68), so the cache never benefits across pods. I'll create a follow-up issue to either:

1. Promote connection caching to the executor level (keyed by `host:port`), or  
2. Explicitly document `GRPCSender` as a single-use, per-`Execute` object and remove the "reuses the connection" language from the struct comment.

Option 1 is the right long-term fix but is a larger refactor; option 2 is the honest short-term fix. The issue will track the connection-pool work.

---

pkg/warmup/grpc_sender.go
hhiroshell: Valid bug. When `resolveMethodDescriptor` fails at line 66–68, `s.conn` is set but `s.methodDesc` remains nil. On every subsequent `Send` call the nil-guard at line 61 passes and a fresh reflection round-trip is attempted, burning a full per-request timeout each time before the loop ends.

Fix: add a `reflectionFailed bool` sentinel field to `GRPCSender`. Set it to `true` on the first reflection error and return `&Response{Error: err}` immediately on later calls without retrying the RPC. The sentinel is cleared on each new dial (when `s.conn == nil`) so a fresh `GRPCSender` always gets one clean attempt.

---

pkg/warmup/grpc_sender.go
hhiroshell: Valid performance issue. `dynamicpb.NewMessage` allocates a new proto message on every `Send` call; with a large `RequestCount` this creates thousands of heap allocations. Since `s.methodDesc` is cached after the first reflection lookup, the input/output descriptors are also stable.

Fix: allocate `reqMsg` (and `respMsg` — see the adjacent comment) once in the struct after the descriptor is resolved, and call `reqMsg.Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool { reqMsg.Clear(fd); return true })` or `proto.Reset(reqMsg)` before each `Unmarshal`. This keeps the object alive and reduces GC pressure proportional to `RequestCount`.

---

pkg/warmup/grpc_sender.go
hhiroshell: Same pattern as the `reqMsg` allocation — `dynamicpb.NewMessage(s.methodDesc.Output())` on every `Send` call is avoidable once the descriptor is cached. Will address both allocations together: cache `reqMsg` and `respMsg` on the struct after the first successful reflection lookup and `Reset()` each before use.

---

pkg/warmup/grpc_sender.go
hhiroshell: Agreed. `methodPath` is `"/" + string(s.methodDesc.Parent().FullName()) + "/" + string(s.methodDesc.Name())` — it's a pure function of `methodDesc` which is immutable after the first reflection lookup. Will cache it as `methodPath string` on the struct alongside `methodDesc`, computed once when the descriptor is resolved.

---

pkg/warmup/grpc_sender.go
hhiroshell: Good catch. Looking at `parseGRPCMethod`:

```go
idx := strings.LastIndex(fullMethod, "/")
if idx < 0 || idx == len(fullMethod)-1 {
```

When input is `"/Check"`, `idx == 0`, so neither guard fires and the function returns `("", "Check", nil)` — an empty `serviceSymbol` with no error. This propagates to `resolveMethodDescriptor` which then fails with a cryptic reflection error.

Fix: add `|| idx == 0` to the guard:

```go
if idx < 0 || idx == 0 || idx == len(fullMethod)-1 {
```

This also needs a test case in `TestParseGRPCMethod` for the leading-slash input `"/Check"` with `wantErr: true`. Currently the test has `{"/Method", "", "Method", false}` which expects no error — that case should be updated to `wantErr: true` after the fix.

---

pkg/warmup/grpc_sender.go
hhiroshell: Valid security concern (CWE-400). `fetchFileDescriptorBytes` returns `fileResp.FileDescriptorResponse.GetFileDescriptorProto()` directly, and `buildFileSet` unmarshals each entry without checking total size. A compromised pod could return a very large or very deep descriptor set and exhaust controller heap, impacting warmup cluster-wide.

Will add a hard limit — a reasonable ceiling is something like 4 MiB total for all descriptor bytes combined before we start unmarshaling. Example:

```go
const maxReflectionResponseBytes = 4 << 20 // 4 MiB

func buildFileSet(fdBytes [][]byte) (*protoregistry.Files, error) {
    var total int
    for _, b := range fdBytes {
        total += len(b)
    }
    if total > maxReflectionResponseBytes {
        return nil, fmt.Errorf("reflection response too large (%d bytes, limit %d)", total, maxReflectionResponseBytes)
    }
    // ... rest of function
}
```

This should be added as a configurable limit (or at minimum a named constant) rather than a magic number.

---

pkg/warmup/grpc_sender.go
hhiroshell: Agreed — `strings.Contains(err.Error(), "already registered")` matches an internal, unstable error string from `google.golang.org/protobuf`. If the upstream wording changes, duplicate-registration errors would be swallowed silently.

The cleaner fix is to check before registering, using `files.FindFileByPath(fdp.GetName())`. If it succeeds, the file is already present and we can skip `RegisterFile` entirely:

```go
if _, err := files.FindFileByPath(fdp.GetName()); err == nil {
    registered[name] = true
    return nil
}
fd, err := protodesc.NewFile(fdp, resolver)
// ...
if err := files.RegisterFile(fd); err != nil {
    return fmt.Errorf("failed to register file descriptor %q: %w", name, err)
}
```

This removes the string-matching hack and makes the duplicate-skip path explicit and verifiable in tests.

---

pkg/warmup/warmup_executor.go
hhiroshell: Agree — `WarmupExecutor` owning an `*http.Client` while the gRPC path instantiates its own `GRPCSender` is dead state today and a smell that will grow as more protocols are added. I'll create a refactoring issue tracking the move to a `Sender` factory or injected `Sender` interface, which also unblocks unit-testing `WarmupExecutor` without real network connections.

---

pkg/warmup/warmup_executor.go
hhiroshell: The concern is valid — creating a new `GRPCSender` (and thus a new gRPC connection) per `Execute` call means N concurrent pod warmups open N connections, even to the same address. The lazy-dial/reflection caching on `GRPCSender` only helps within a single pod's request loop.

However, implementing a proper connection pool (keyed by `host:port`) and a `Sender` factory interface is a non-trivial refactor that would significantly expand the scope of this PR. Given the PR is already large (gRPC support + reflection + executor refactor), I'd prefer to:

1. Merge this PR to ship gRPC warmup support.
2. Track the connection pooling + `Sender` factory work in a dedicated follow-up issue.

The testability concern (no mock injection) is legitimate and will be addressed in the same follow-up. Current tests exercise `GRPCSender` directly with a real gRPC server, which provides reasonable coverage in the meantime.

---

pkg/warmup/warmup_executor.go
hhiroshell: Valid defensive-coding issue. The current `default:` silently falls back to HTTP for any unknown protocol, which would mask bugs when a new protocol is added but its case is forgotten.

Will change to:

```go
case ProtocolHTTP:
    sender = &HTTPSender{client: e.client, logger: e.logger}
    // ...
default:
    return &Result{
        Error:   fmt.Errorf("unsupported warmup protocol %q", config.Protocol),
        Message: fmt.Sprintf("cannot execute warmup: unsupported protocol %q", config.Protocol),
    }
```

This mirrors `ParseConfig` which already validates the protocol at parse time, so `default` should theoretically never be reached — but it's the right safety net to have.

---

pkg/warmup/warmup_executor.go
hhiroshell: Good catch. Looking at commit 0ca30b3, the original comment explained that unattempted requests (those cut short by a break) are counted as failed, so the final `Result` message and the fail-open decision in the controller reflect the complete `RequestCount` picture rather than just the requests that ran.

Will restore the fuller comment:

```go
// Count remaining un-attempted requests as failed so the result message and
// the fail-open decision in the controller reflect the full RequestCount picture.
failCount += config.RequestCount - successCount - failCount
```

---

pkg/warmup/warmup_executor.go
hhiroshell: The current behavior is intentional for the ASAP model: all `RequestCount` iterations are attempted regardless of per-request errors, only breaking on context cancellation/deadline. The idea is to send as many warmup requests as possible within the configured timeout.

That said, the reflection-failure scenario is a good edge case: if `resolveMethodDescriptor` fails, the sentinel fix planned in [the reflection lookup thread] means subsequent `Send` calls return immediately after the sentinel is set rather than retrying a full reflection round-trip. So once the sentinel is in place, the looping behavior on a permanently-failing reflection endpoint is cheap (near-instant returns), not expensive.

For genuinely permanent transport errors (e.g. pod IP unreachable), your point about error structs distinguishing permanent vs transient errors is interesting. A `PermanentError` type that signals "break immediately" vs the current `continue` behavior would be a clean extension. I'd propose tracking that as a follow-up issue rather than adding it here, since the current semantics are documented and the reflection case will be addressed by the sentinel.

---

pkg/warmup/config.go
hhiroshell: Will create a follow-up issue. The flat `Config` struct is workable for two protocols, but embedding `HTTPConfig` / `GRPCConfig` sub-structs would avoid the sparsely-populated pattern and make it clear which fields apply to which protocol. The issue will also cover updating `ParseConfig` to populate sub-structs rather than flat fields.

---

pkg/warmup/config.go
hhiroshell: Agree — the `GRPCMethod` annotation is stored verbatim and only checked for emptiness (line 145). A value like `"/Check"` or `"NoSlash"` passes `ParseConfig` and only fails at runtime when `GRPCSender.Send` calls `parseGRPCMethod`.

Fix: call `parseGRPCMethod` in `ParseConfig` after storing the method and return a descriptive error if it's malformed:

```go
if method, ok := annotations[webhook.AnnotationWarmupGRPCMethod]; ok && method != "" {
    if _, _, err := parseGRPCMethod(method); err != nil {
        return config, fmt.Errorf("invalid %s value: %w", webhook.AnnotationWarmupGRPCMethod, err)
    }
    config.GRPCMethod = method
}
```

This surfaces misconfiguration at annotation parse time (with a clear error in the controller event) instead of at warmup execution time.

---

pkg/warmup/config.go
hhiroshell: Webhook-level validation is the right direction — it surfaces annotation errors at admission time before the pod is even scheduled, giving immediate feedback via the `kubectl apply` response.

The challenge is that full payload validation requires the proto descriptor for the target method, which the webhook doesn't have without performing a gRPC reflection call itself (and the target pod may not be running yet at admission time).

A practical split:

- **In `ParseConfig` now**: validate JSON syntax with `json.Valid([]byte(payload))`. This catches formatting errors early and is cheap to add.
- **In the webhook (follow-up)**: at minimum, validate JSON syntax during admission. Proto-schema validation against the actual message type is only possible at warmup time.

I'll add the JSON syntax check in `ParseConfig` in this PR and open an issue for the webhook-level annotation validation.

---

pkg/warmup/sender.go
hhiroshell: Agreed — the shared `Target` struct with protocol-conditional fields can be confusing. Will add explicit "ignored by" notes to the field comments:

```go
// Headers contains additional request headers. Ignored by GRPCSender.
Headers map[string]string

// Payload is the request body; JSON-encoded for gRPC, empty for HTTP GET.
// Ignored by HTTPSender (which uses no body for GET requests).
Payload []byte
```

This makes the cross-sender contract visible without changing the struct shape.

---

pkg/warmup/grpc_sender_test.go
hhiroshell: Agree — the `var _ = fmt.Sprintf` suppression is misleading: it looks like an intentional API anchor rather than a workaround for an unused import. The `fmt` package isn't used anywhere else in `grpc_sender_test.go`. Will remove both the import and the suppression line.

---

pkg/warmup/config_test.go
hhiroshell: Valid testing gap. The `!= ""` guards mean that every pre-existing test case with a zero-value `wantConfig` silently skips the `Protocol` check, so none of them verify that `ParseConfig` correctly defaults to `ProtocolHTTP`.

Fix: make all three checks unconditional and ensure every test case explicitly sets `wantConfig.Protocol`. Pre-existing cases should have `Protocol: ProtocolHTTP` added to their `wantConfig`. This also makes it impossible to accidentally add a case that asserts "protocol must be empty", which is never a valid expectation.

---

cmd/controller/main.go
hhiroshell: Agreed — at this stage of the project there are no known external consumers and the rename correctly reflects the protocol-agnostic scope of the executor. No action needed beyond the existing call-site updates in this PR.

---

config/samples/sample_grpc_deployment.yaml
hhiroshell: Agreed. The health-check endpoint is used here because it requires no custom proto and works out-of-the-box with the sample server. Will add a comment making that explicit:

```yaml
# The health-check endpoint is used here for demonstration (no custom proto required).
# In practice, use your application's actual warmup endpoint (e.g., a cache-preload RPC).
kube-booster.io/warmup-grpc-method: "grpc.health.v1.Health/Check"
```

This makes the intent clear without changing the sample's functional behavior.

---

pkg/warmup/warmup_executor.go
hhiroshell: Agree. The two separate `V(1).Info("starting ... warmup", ...)` calls duplicate the common fields (`pod`, `namespace`, `requestCount`, `timeout`) and diverge only in protocol-specific fields. Since `config.Protocol` is available after the `switch`, the log can be moved out:

```go
// after the switch block, before defer sender.Close()
e.logger.V(1).Info("starting warmup",
    "pod", config.PodName,
    "namespace", config.PodNamespace,
    "protocol", config.Protocol,
    "address", target.Address,
    "method", target.Method,
    "requestCount", config.RequestCount,
    "timeout", config.Timeout)
```

This removes the duplication and ensures the log line is always emitted regardless of which protocol branch runs.

---

pkg/warmup/http_sender_test.go
hhiroshell: Agreed. The `kube-booster/1.0` string in the `User-Agent` header (and this test assertion) should come from a build-time variable or a shared package-level constant rather than being hardcoded. 

The typical Go pattern is a package-level `var Version = "dev"` in a `version` package, overridden at link time with `-ldflags "-X .../version.Version=v1.2.3"`. The `User-Agent` header and the test assertion would then reference `version.Version`. Will track this in the existing build/versioning issue (or open one if there isn't one).

---

pkg/warmup/grpc_sender.go
hhiroshell: The `Duration` in the error response at line 64 (`parseGRPCMethod` failure) is near-zero since it's just local string parsing — not a real round-trip. For the reflection failure case at line 68, it does represent time spent communicating with the server, which is meaningful for timeout accounting.

However, looking at how `WarmupExecutor.Execute` uses the response: it only records latency when `resp.Error == nil` (line 133 in `warmup_executor.go`), so this `Duration` is discarded in both cases. The field is included for consistency with the `Response` type contract, but it isn't observable to callers today.

Options:
1. Keep it as-is (consistent but unused in the error path).
2. Drop `Duration` from error responses (`return &Response{Error: err}`) to signal "no meaningful timing" — but this breaks the type contract symmetry.
3. Clarify in the `Response.Duration` doc comment that it may be zero or approximate when `Error != nil`.

Option 3 seems cleanest for now. Will update the comment.

---

pkg/warmup/grpc_sender.go
hhiroshell: Agreed — the unary-only constraint is important for users to know before they configure `warmup-grpc-method`. Will add it to `docs/USAGE.md` under the gRPC warmup section, e.g.:

> **Note:** kube-booster only supports unary RPCs. Streaming methods (client-streaming, server-streaming, or bidirectional) are rejected at warmup time with an error event. Use a unary RPC such as a health-check or a lightweight read-only call.

Also worth adding a note in the annotation reference in `CLAUDE.md` / `docs/USAGE.md` next to the `warmup-grpc-method` annotation.

---

pkg/warmup/grpc_sender_test.go
hhiroshell: The test currently verifies connection reuse implicitly — three `Send` calls all succeed, and since `GRPCSender` only calls `grpc.NewClient` when `s.conn == nil`, if the second and third calls succeed we know the connection was reused (a broken reuse would fail the dial on the second call).

To verify more explicitly, the options are:

1. **Inspect the struct field** — `conn` is unexported, but since the test is in the same package (`package warmup`), it can access `sender.conn` directly. We can assert `sender.conn != nil` after the first `Send` and store the pointer, then check it's the same pointer after the second and third calls.

2. **Count server-side connections** — the test gRPC server could record the number of accepted connections and assert it's 1.

Option 1 is simpler and sufficient. Will update the test:

```go
// After first Send, connection should be established.
if sender.conn == nil {
    t.Fatal("expected conn to be non-nil after first Send")
}
firstConn := sender.conn
// ...subsequent sends...
if sender.conn != firstConn {
    t.Error("expected same connection to be reused across Send calls")
}
```

---

pkg/warmup/grpc_sender.go
hhiroshell: Let's go with the option 1.

pkg/warmup/grpc_sender.go
hhiroshell: **Create issue for optional TLS**: new issue

Created issue #60 to track adding `--grpc-tls-cert` / `--grpc-tls-key` flags and injecting TLS credentials into `GRPCSender`. https://github.com/hhiroshell/kube-booster/issues/60

**Document the plaintext-only constraint**: fix

Fixed in commit f6c2f3e. Added a dedicated "gRPC Warmup" section to `docs/USAGE.md` that documents the plaintext transport constraint, states that pod-to-pod traffic is commonly treated as trusted, and recommends a NetworkPolicy to restrict controller-to-pod traffic while TLS support is pending.

---

pkg/warmup/grpc_sender.go
hhiroshell: **Update GRPCSender doc comment to reflect single-use scope**: fix

Fixed in commit b3eb520. The struct comment now explicitly states "single-use, per-Execute object" with no cross-pod connection reuse, and the new fields (`methodPath`, `reqMsg`, `respMsg`) are annotated accordingly.

**Create issue for cross-pod connection pooling**: new issue

Created issue #61 to track promoting connection caching to the executor level (keyed by `host:port`) and the Sender factory refactoring that enables it. https://github.com/hhiroshell/kube-booster/issues/61

---

pkg/warmup/grpc_sender.go
hhiroshell: **Add reflectionFailed sentinel**: fix

Fixed in commit b3eb520. Added a `reflectionFailed bool` field to `GRPCSender`. It is set to `true` on the first parse error or reflection failure and short-circuits subsequent `Send` calls with an immediate error return, preventing redundant network round-trips on a permanently-failing reflection endpoint.

---

pkg/warmup/grpc_sender.go
hhiroshell: **Cache reqMsg and reset before each use**: fix

Fixed in commit b3eb520. `reqMsg` and `respMsg` are now allocated once when the method descriptor is first resolved and stored as struct fields. `proto.Reset` is called before each use, eliminating the per-`Send` allocation.

---

pkg/warmup/grpc_sender.go
hhiroshell: **Cache respMsg and reset before each use**: fix

Fixed in commit b3eb520 together with `reqMsg`. Both messages are allocated once after the descriptor is resolved and reused with `proto.Reset` per call.

---

pkg/warmup/grpc_sender.go
hhiroshell: **Cache methodPath on struct**: fix

Fixed in commit b3eb520. `methodPath` is computed once when the descriptor is resolved and stored as `s.methodPath`, eliminating the string concatenation on every `Send` call.

---

pkg/warmup/grpc_sender.go
hhiroshell: **Reject leading-slash methods in parseGRPCMethod**: fix

Fixed in commit b3eb520. Added `|| idx == 0` to the guard so `"/Check"` now returns an error rather than an empty `serviceSymbol`.

**Update TestParseGRPCMethod for the leading-slash case**: fix

Fixed in commit 7a17d46. Changed the `{"/Method", "", "Method", false}` test case to `wantErr: true` to match the corrected behavior.

---

pkg/warmup/grpc_sender.go
hhiroshell: **Add 4 MiB size cap in buildFileSet**: fix

Fixed in commit b3eb520. `buildFileSet` now sums the total byte count of all `FileDescriptorProto` entries before any unmarshaling and returns an error if the total exceeds 4 MiB (`maxReflectionResponseBytes = 4 << 20`). This caps memory exposure from a compromised pod's reflection response (CWE-400).

---

pkg/warmup/grpc_sender.go
hhiroshell: **Use FindFileByPath before RegisterFile**: fix

Fixed in commit b3eb520. The `register` closure now calls `files.FindFileByPath(fdp.GetName())` before constructing a new descriptor. If the file is already present, it marks it as registered and returns early — removing the `strings.Contains(err.Error(), "already registered")` string-matching guard entirely.

---

pkg/warmup/warmup_executor.go
hhiroshell: **Create refactoring issue for WarmupExecutor Sender factory**: new issue

Created issue #61 to track moving to a `Sender` factory / injected sender interface, which removes the dead `*http.Client` field and unblocks unit-testing `WarmupExecutor` without real network connections. The issue also covers cross-pod connection pooling. https://github.com/hhiroshell/kube-booster/issues/61

---

pkg/warmup/warmup_executor.go
hhiroshell: **Create issue for connection pooling and Sender factory**: new issue

Deferring to a follow-up as agreed. Created issue #61 covering: (1) cross-pod connection pooling keyed by `host:port`, (2) `Sender` factory interface to enable pooling and testability, (3) removal of the dead `*http.Client` field. https://github.com/hhiroshell/kube-booster/issues/61

---

pkg/warmup/warmup_executor.go
hhiroshell: **Use explicit case ProtocolHTTP in switch**: fix

Fixed in commit d71dd88. Changed `default:` to `case ProtocolHTTP:` and added a new `default:` that returns an error for unrecognised protocols, preventing silent fallback to HTTP if a new protocol case is forgotten.

---

pkg/warmup/warmup_executor.go
hhiroshell: **Restore fuller fail-count comment**: fix

Fixed in commit d71dd88. Restored the explanation that un-attempted requests are counted as failed so the result message and fail-open decision in the controller reflect the full `RequestCount` picture, not just the requests that ran before the break.

---

pkg/warmup/warmup_executor.go
hhiroshell: **Loop behavior on persistent errors**: won't fix

No action taken: the ASAP warmup model intentionally retries all `RequestCount` iterations, breaking only on context cancellation/deadline. The reflection-failure case (the most expensive scenario discussed) is now handled by the `reflectionFailed` sentinel added in commit b3eb520 — subsequent `Send` calls return immediately after the first failure, so the loop is cheap even when reflection is permanently unavailable.

**PermanentError type for genuine transport failures**: new issue

Created issue #62 to track a `PermanentError` sentinel type that senders can use to signal "break the loop immediately" for non-recoverable conditions (e.g. unreachable pod IP, authentication failure). https://github.com/hhiroshell/kube-booster/issues/62

---

pkg/warmup/config.go
hhiroshell: **Create issue for Config protocol sub-structs**: new issue

Created issue #63 to track refactoring `Config` to embed `HTTPConfig` / `GRPCConfig` sub-structs, which would eliminate the sparsely-populated flat struct pattern and make protocol-specific fields self-documenting. https://github.com/hhiroshell/kube-booster/issues/63

---

pkg/warmup/config.go
hhiroshell: **Validate GRPCMethod format in ParseConfig**: fix

Fixed in commit 6af1244. `ParseConfig` now calls `parseGRPCMethod` when storing the `warmup-grpc-method` annotation. A malformed value such as `"/Check"` or `"NoSlash"` returns a descriptive error at annotation parse time rather than at warmup execution time. A corresponding test case was added in the same commit.

---

pkg/warmup/config.go
hhiroshell: **JSON syntax validation in ParseConfig**: fix

Fixed in commit 6af1244. `ParseConfig` now calls `json.Valid` on the `warmup-grpc-payload` annotation before storing it. Malformed JSON is rejected with a descriptive error at parse time rather than at warmup time when `protojson.Unmarshal` would otherwise fail silently (fail-open). A test case for invalid JSON payload was added.

**Webhook-level annotation validation**: new issue

Created issue #64 to track moving JSON syntax and method format checks into the admission webhook so users see errors at `kubectl apply` time. Proto-schema validation against the actual message type can only be done at warmup time and is out of scope for the webhook. https://github.com/hhiroshell/kube-booster/issues/64

---

pkg/warmup/sender.go
hhiroshell: **Add 'ignored by' notes to Target field comments**: fix

Fixed in commit 2afbb3c. Updated the `Headers` and `Payload` field comments to say "Ignored by GRPCSender" and "Ignored by HTTPSender" respectively, making the cross-sender contract explicit without changing the struct shape.

---

pkg/warmup/grpc_sender_test.go
hhiroshell: **Remove unused fmt import and suppression line**: fix

Fixed in commit 7a17d46. Removed the `"fmt"` import and the `var _ = fmt.Sprintf` suppression line from `grpc_sender_test.go`.

---

pkg/warmup/config_test.go
hhiroshell: **Make Protocol/GRPCMethod/GRPCPayload assertions unconditional**: fix

Fixed in commit 09aa141. Removed the `!= ""` guards on all three assertions. Added `Protocol: ProtocolHTTP` and `GRPCPayload: DefaultGRPCPayload` to six pre-existing `wantConfig` structs that omitted them. Added two new test cases for invalid GRPCMethod format and invalid JSON payload to cover the new `ParseConfig` validation.

---

config/samples/sample_grpc_deployment.yaml
hhiroshell: **Clarify health-check is for demonstration only**: fix

Fixed in commit 5fc8a18. Added a comment to the sample YAML explaining that the health-check gRPC method is used because it requires no custom proto. Users should replace it with their application's actual warmup endpoint.

---

pkg/warmup/warmup_executor.go
hhiroshell: **Unify log statement after switch**: fix

Fixed in commit d71dd88. Removed the two per-protocol `V(1).Info` calls and added a single unified `"starting warmup"` log after the switch block. It includes `protocol`, `address`, and `method` fields so the log is equally informative for both HTTP and gRPC.

---

pkg/warmup/http_sender_test.go
hhiroshell: **Create issue for version-aware User-Agent**: new issue

Created issue #65 to track introducing a `pkg/version` package with a build-time injectable `Version` variable. The `User-Agent` header and test assertion would reference `version.Version` instead of the hard-coded `"1.0"`. https://github.com/hhiroshell/kube-booster/issues/65

---

pkg/warmup/grpc_sender.go
hhiroshell: **Keep Duration in error responses as-is**: won't fix

No action taken: applying the user's choice of option 1. The `Duration` field is included in error responses for type-contract consistency; its value is not used by `WarmupExecutor` (which only records latency when `resp.Error == nil`). Keeping it as-is avoids a contract asymmetry without any observable downside.

---

pkg/warmup/grpc_sender.go
hhiroshell: **Document unary-only constraint in docs/USAGE.md**: fix

Fixed in commit f6c2f3e. Added a "gRPC Warmup" section to `docs/USAGE.md` that explicitly states only unary RPCs are supported, that streaming methods are rejected with a `WarmupFailed` event, and recommends using a unary health-check or lightweight read-only call.

---

pkg/warmup/grpc_sender_test.go
hhiroshell: **Explicitly verify conn pointer reuse**: fix

Fixed in commit 7a17d46. `TestGRPCSender_Send_ConnectionReuse` now captures `sender.conn` after the first `Send` and asserts the pointer is the same after subsequent calls, making the connection-reuse guarantee explicit rather than implicit.

---

pkg/warmup/grpc_sender.go
hhiroshell: Please create a follow-up issue to make `maxReflectionResponseBytes` configurable.

hhiroshell: **Make maxReflectionResponseBytes configurable**: new issue

Created issue #66 to track exposing the 4 MiB limit as a controller flag (e.g. `--grpc-reflection-max-response-bytes`), passed down through `WarmupExecutor` → `GRPCSender` so operators can raise or lower it without recompiling. The named constant remains as the default. https://github.com/hhiroshell/kube-booster/issues/66

---

PR #70, opened 2026-06-25
feat: add WarmupConfig CRD for scenario-based warmup (#26)

### LINKED ISSUE #26: Add scenario-based warmup execution with WarmupConfig CRD
## Context

Simple single-endpoint warmup is insufficient for complex applications with interdependent services, internal caches, and JIT compilation paths. Applications need multi-step warmup scenarios: authenticate first, then load cache, then verify health — with response data chaining between steps.

This phase introduces a `WarmupConfig` CRD for declarative scenario definitions and a scenario executor that orchestrates sequential/parallel steps with response extraction.

This is **Phase 4** of the warmup engine replacement.
- Phase 1: #23 — Replace Vegeta with custom HTTP executor
- Phase 2: #24 — Cross-pod concurrency control
- Phase 3: #25 — gRPC support

**Depends on:** #25

### Example WarmupConfig

```yaml
apiVersion: kube-booster.io/v1alpha1
kind: WarmupConfig
metadata:
  name: my-app-warmup
spec:
  timeout: "120s"
  steps:
    - name: "authenticate"
      requests:
        - endpoint: /api/auth/token
          method: POST
          body: '{"client": "warmup"}'
          extract:
            token: "$.access_token"

    - name: "warm cache"
      requests:
        - endpoint: /api/cache/load
          method: POST
          headers:
            Authorization: "Bearer {{token}}"
          body: '{"action": "preload"}'
          count: 3
          extract:
            cacheId: "$.id"

    - name: "verify endpoints"
      parallel: true
      requests:
        - endpoint: /api/v1/health
          expectedStatus: 200
        - endpoint: /api/v2/health
          expectedStatus: 200

    - name: "warm gRPC"
      requests:
        - protocol: grpc
          grpcMethod: "myapp.Cache/Warm"
          grpcPayload: '{"key": "{{cacheId}}"}'
          count: 5
```

### Pod Annotation

```yaml
annotations:
  kube-booster.io/warmup: "enabled"
  kube-booster.io/warmup-config: "my-app-warmup"    # References the CRD above
  kube-booster.io/warmup-port: "8080"
```

When `warmup-config` annotation is present, the CRD-based scenario takes precedence over individual annotation-based configuration.

### Step Execution Model

```
Scenario
├── Step 1: [Request A] → extract token         ← sequential
├── Step 2: [Request B x3] → extract cacheId    ← sequential, repeated
├── Step 3: [Request C, Request D]               ← parallel (errgroup)
└── Step 4: [gRPC Request E x5]                  ← sequential, repeated
```

- Steps execute **sequentially** (step N must complete before step N+1)
- Requests within a step: **parallel** if `parallel: true`, otherwise **sequential**
- Each request can be repeated `count` times (back-to-back ASAP)
- Response extraction stores values in a **session context**
- `{{varName}}` placeholders in subsequent requests are interpolated from session

### ScenarioExecutor

```go
type ScenarioExecutor struct {
    logger      logr.Logger
    rateLimiter *RequestRateLimiter
}

func (se *ScenarioExecutor) ExecuteScenario(
    ctx context.Context, config *Config, scenario *v1alpha1.WarmupConfigSpec,
) *Result
```

### SessionContext

```go
type SessionContext struct { ... }
func (sc *SessionContext) Set(key string, value interface{})
func (sc *SessionContext) Get(key string) (interface{}, bool)
func (sc *SessionContext) Interpolate(s string) string  // replaces {{key}} placeholders
```

Thread-safe (uses `sync.RWMutex`) for parallel step execution.

### Controller Dispatch

```go
// In Reconcile:
if config.WarmupConfigName != "" {
    warmupConfig := &v1alpha1.WarmupConfig{}
    err := r.Get(ctx, types.NamespacedName{Name: config.WarmupConfigName, Namespace: pod.Namespace}, warmupConfig)
    // Use ScenarioExecutor
    result = scenarioExecutor.ExecuteScenario(warmupCtx, config, &warmupConfig.Spec)
} else {
    // Use default WarmupExecutor (existing annotation-based behavior)
    result = r.WarmupExecutor.Execute(warmupCtx, config)
}
```

## Files to Change

| Action | File | What |
|--------|------|------|
| Create | `config/crd/warmupconfig.yaml` | CRD definition |
| Create | `pkg/api/v1alpha1/types.go` | Go types: `WarmupConfig`, `WarmupConfigSpec`, `WarmupStep`, `WarmupRequest` |
| Create | `pkg/api/v1alpha1/register.go` | Scheme registration |
| Create | `pkg/api/v1alpha1/deepcopy.go` | Hand-written DeepCopy methods |
| Create | `pkg/warmup/scenario_executor.go` | `ScenarioExecutor` orchestrating steps |
| Create | `pkg/warmup/scenario_executor_test.go` | Tests |
| Create | `pkg/warmup/session.go` | `SessionContext` for response chaining |
| Create | `pkg/warmup/session_test.go` | Tests |
| Modify | `pkg/webhook/constants.go` | Add `AnnotationWarmupConfig` constant |
| Modify | `pkg/warmup/config.go` | Add `WarmupConfigName` field; parse annotation |
| Modify | `pkg/controller/pod_controller.go` | Add CRD lookup; dispatch to ScenarioExecutor |
| Modify | `cmd/controller/main.go` | Register v1alpha1 scheme |
| Modify | `config/rbac/role.yaml` | Add `warmupconfigs` get/list/watch permissions |
| Modify | `config/kustomization.yaml` | Add CRD resource |

## WarmupConfigSpec Go Types

```go
type WarmupConfigSpec struct {
    Steps   []WarmupStep `json:"steps"`
    Timeout string       `json:"timeout,omitempty"`  // default: "120s"
}

type WarmupStep struct {
    Name     string          `json:"name,omitempty"`
    Parallel bool            `json:"parallel,omitempty"`   // default: false
    Requests []WarmupRequest `json:"requests"`
    Timeout  string          `json:"timeout,omitempty"`    // default: "30s"
}

type WarmupRequest struct {
    Name           string            `json:"name,omitempty"`
    Protocol       string            `json:"protocol,omitempty"`      // "http" or "grpc"
    Endpoint       string            `json:"endpoint"`
    Method         string            `json:"method,omitempty"`        // HTTP method, default: "GET"
    Headers        map[string]string `json:"headers,omitempty"`
    Body           string            `json:"body,omitempty"`
    GRPCMethod     string            `json:"grpcMethod,omitempty"`
    GRPCPayload    string            `json:"grpcPayload,omitempty"`
    Count          int               `json:"count,omitempty"`         // default: 1
    Extract        map[string]string `json:"extract,omitempty"`       // key: varName, value: JSONPath
    ExpectedStatus int               `json:"expectedStatus,omitempty"`
}
```

## Edge Cases and Error Handling

- **CRD not found** → fail-open with warning event
- **Both CRD ref and annotations** → CRD takes precedence
- **Empty steps array** → return success immediately
- **Step timeout exceeded** → cancel remaining requests in step, continue to next step
- **Overall timeout exceeded** → cancel all remaining steps, return partial results
- **Extract failure** → log warning, continue (don't fail warmup)
- **Interpolation of missing variable** → leave `{{varName}}` as-is, log warning
- **Parallel step with same extract key** → last-write-wins (non-deterministic); documented limitation

## Tests

- CRD types: serialize/deserialize round-trip
- SessionContext: concurrent Set/Get, Interpolate with placeholders and missing vars
- ScenarioExecutor: sequential steps, parallel steps, response chaining, step timeout, overall timeout
- Controller: CRD lookup from fake client, fallback to annotation config when CRD not found

## Verification

```bash
make test && make lint && make build
kubectl apply --dry-run=client -f config/crd/warmupconfig.yaml  # CRD valid
```

## Documentation Updates

- `docs/USAGE.md`: Add WarmupConfig CRD section with complete examples
- `docs/DEVELOPMENT.md`: Document CRD types and scenario executor
- `docs/IMPLEMENTATION_SUMMARY.md`: Update from "planned" to "implemented"
- `CLAUDE.md`: Update CRD section from "Future" to "Implemented"
- `config/samples/sample_warmup_config.yaml`: Example CRD
- `config/samples/sample_scenario_deployment.yaml`: Example deployment referencing CRD

hhiroshell:
## Summary

Implements #26.

Adds scenario-based warmup execution driven by a `WarmupConfig` custom resource. When a pod carries the `kube-booster.io/warmup-config` annotation, the controller fetches the named CRD and uses the new `ScenarioExecutor` to run multi-step warmup with response chaining, instead of the single-endpoint annotation-based path.

- **`pkg/api/v1alpha1/`** — `WarmupConfig`, `WarmupStep`, `WarmupRequest` Go types with hand-written `DeepCopy*` and scheme registration
- **`config/crd/warmupconfig.yaml`** — CRD manifest with OpenAPI v3 schema
- **`pkg/warmup/session.go`** — thread-safe `SessionContext` for `{{varName}}` interpolation between requests
- **`pkg/warmup/scenario_executor.go`** — `ScenarioExecutor` orchestrating steps sequentially with per-step timeouts, JSON response extraction (`$.key`, `$.a.b`), and shared rate limiter
- **`pkg/warmup/http_sender.go`** — extended to support arbitrary HTTP method and request body (needed for scenario POST/PUT requests); response body now captured for extraction
- **Controller dispatch** — `PodReconciler` dispatches to `ScenarioExecutor` when `WarmupConfigName` is set; falls back to annotation-based warmup with a `Warning` event if the CRD is not found (fail-open)
- **RBAC** — added `get/list/watch` for `warmupconfigs` to `config/rbac/role.yaml`
- **Kustomize** — `config/crd/warmupconfig.yaml` added to `resources:`
- **Sample files** — `config/samples/sample_warmup_config.yaml` and `config/samples/sample_scenario_deployment.yaml`
- **Docs** — `docs/USAGE.md` (WarmupConfig CRD section), `docs/DEVELOPMENT.md`, `docs/IMPLEMENTATION_SUMMARY.md`, `CLAUDE.md`

Note: `parallel` step execution is intentionally absent from this PR (tracked in #69).

## Changes Made

- 12 new files (3 Go types, 2 executors+tests, 1 session+test, 1 CRD manifest, 2 samples, 1 Kustomize update)
- 13 modified files (controller, config, webhook constants, warmup config, HTTP sender, mock, docs)
- 1,761 insertions, 51 deletions

## Testing

- All existing tests continue to pass (`go test ./...`)
- New tests: `session_test.go`, `scenario_executor_test.go`, controller dispatch tests in `pod_controller_test.go`
- `make lint` passes (0 issues)
- `make build` succeeds

## Related Issue

Closes #26

---
*PR generated with [Claude Code](https://claude.ai/code) - github-devflow:implement*

### DIFF
--- a/config/crd/warmupconfig.yaml
+++ b/config/crd/warmupconfig.yaml
@@ -31,7 +31,8 @@ spec:
               properties:
                 timeout:
                   type: string
-                  description: "Overall time limit for all steps (Go duration, e.g. '120s'). Default: '120s'."
+                  maxLength: 32
+                  description: "Overall time limit for all steps (Go duration, e.g. '120s'). Default: '120s'. Capped at 5m."
                 steps:
                   type: array
                   minItems: 1
@@ -42,9 +43,11 @@ spec:
                     properties:
                       name:
                         type: string
+                        maxLength: 1024
                         description: "Optional human-readable label for log output."
                       timeout:
                         type: string
+                        maxLength: 32
                         description: "Time limit for this step (Go duration). Default: '30s'."
                       requests:
                         type: array
@@ -55,6 +58,7 @@ spec:
                           properties:
                             name:
                               type: string
+                              maxLength: 1024
                               description: "Optional label for log output."
                             protocol:
                               type: string
@@ -62,32 +66,42 @@ spec:
                               description: "Warmup transport protocol. Default: 'http'."
                             endpoint:
                               type: string
+                              maxLength: 1024
                               description: "URL path for HTTP requests (e.g. '/api/warmup'). Ignored for gRPC."
                             method:
                               type: string
+                              maxLength: 16
                               description: "HTTP verb (e.g. 'GET', 'POST'). Default: 'GET'. Ignored for gRPC."
                             headers:
                               type: object
+                              maxProperties: 50
                               additionalProperties:
                                 type: string
+                                maxLength: 4096
                               description: "Additional HTTP request headers. Supports {{varName}} interpolation."
                             body:
                               type: string
+                              maxLength: 65536
                               description: "HTTP request body. Supports {{varName}} interpolation. Ignored for gRPC."
                             grpcMethod:
                               type: string
+                              maxLength: 1024
                               description: "Fully-qualified gRPC method ('pkg.Service/Method'). Required when protocol is 'grpc'."
                             grpcPayload:
                               type: string
+                              maxLength: 65536
                               description: "JSON-encoded gRPC request message. Supports {{varName}} interpolation. Default: '{}'."
                             count:
                               type: integer
                               minimum: 1
+                              maximum: 12000
                               description: "Number of times to repeat this request. Default: 1."
                             extract:
                               type: object
+                              maxProperties: 50
                               additionalProperties:
                                 type: string
+                                maxLength: 1024
                               description: "Map from session variable name to JSONPath expression ($.key or $.a.b). Extracted from last response body."
                             expectedStatus:
                               type: integer
--- a/config/crd/warmupconfig.yaml
+++ b/config/crd/warmupconfig.yaml
@@ -1,0 +1,94 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: warmupconfigs.kube-booster.io
+spec:
+  group: kube-booster.io
+  names:
+    kind: WarmupConfig
+    listKind: WarmupConfigList
+    plural: warmupconfigs
+    singular: warmupconfig
+  scope: Namespaced
+  versions:
+    - name: v1alpha1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          required: ["spec"]
+          properties:
+            apiVersion:
+              type: string
+            kind:
+              type: string
+            metadata:
+              type: object
+            spec:
+              type: object
+              required: ["steps"]
+              properties:
+                timeout:
+                  type: string
+                  description: "Overall time limit for all steps (Go duration, e.g. '120s'). Default: '120s'."
+                steps:
+                  type: array
+                  minItems: 1
+                  description: "Ordered list of warmup steps executed sequentially."
+                  items:
+                    type: object
+                    required: ["requests"]
+                    properties:
+                      name:
+                        type: string
+                        description: "Optional human-readable label for log output."
+                      timeout:
+                        type: string
+                        description: "Time limit for this step (Go duration). Default: '30s'."
+                      requests:
+                        type: array
+                        minItems: 1
+                        description: "Warmup requests executed sequentially within this step."
+                        items:
+                          type: object
+                          properties:
+                            name:
+                              type: string
+                              description: "Optional label for log output."
+                            protocol:
+                              type: string
+                              enum: ["http", "grpc"]
+                              description: "Warmup transport protocol. Default: 'http'."
+                            endpoint:
+                              type: string
+                              description: "URL path for HTTP requests (e.g. '/api/warmup'). Ignored for gRPC."
+                            method:
+                              type: string
+                              description: "HTTP verb (e.g. 'GET', 'POST'). Default: 'GET'. Ignored for gRPC."
+                            headers:
+                              type: object
+                              additionalProperties:
+                                type: string
+                              description: "Additional HTTP request headers. Supports {{varName}} interpolation."
+                            body:
+                              type: string
+                              description: "HTTP request body. Supports {{varName}} interpolation. Ignored for gRPC."
+                            grpcMethod:
+                              type: string
+                              description: "Fully-qualified gRPC method ('pkg.Service/Method'). Required when protocol is 'grpc'."
+                            grpcPayload:
+                              type: string
+                              description: "JSON-encoded gRPC request message. Supports {{varName}} interpolation. Default: '{}'."
+                            count:
+                              type: integer
+                              minimum: 1
+                              description: "Number of times to repeat this request. Default: 1."
+                            extract:
+                              type: object
+                              additionalProperties:
+                                type: string
+                              description: "Map from session variable name to JSONPath expression ($.key or $.a.b). Extracted from last response body."
+                            expectedStatus:
+                              type: integer
+                              description: "HTTP status code that counts as success. When 0, 200–399 are success. Ignored for gRPC."
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -224,6 +224,11 @@ kube-booster/
 │   └── controller/
 │       └── main.go              # Main entry point
 ├── pkg/
+│   ├── api/
+│   │   └── v1alpha1/
+│   │       ├── types.go          # WarmupConfig, WarmupStep, WarmupRequest Go types
+│   │       ├── register.go       # Scheme registration (AddToScheme)
+│   │       └── deepcopy.go       # Hand-written DeepCopy* methods
 │   ├── controller/
 │   │   ├── pod_controller.go     # Reconciler implementation
 │   │   ├── pod_controller_test.go
@@ -236,12 +241,16 @@ kube-booster/
 │   │   ├── config_test.go
 │   │   ├── grpc_sender.go        # GRPCSender: gRPC warmup via server reflection
 │   │   ├── grpc_sender_test.go
-│   │   ├── http_sender.go        # HTTPSender: single HTTP GET request
+│   │   ├── http_sender.go        # HTTPSender: HTTP warmup (GET/POST/etc. + body)
 │   │   ├── http_sender_test.go
-│   │   ├── mock.go               # MockSender for testing
+│   │   ├── mock.go               # MockExecutor / MockScenarioExecutor for testing
 │   │   ├── rate_limiter.go       # Nil-safe RPS rate limiter wrapper
 │   │   ├── result.go             # Warmup result structure
+│   │   ├── scenario_executor.go  # ScenarioExecutor: multi-step CRD-based warmup
+│   │   ├── scenario_executor_test.go
 │   │   ├── sender.go             # Sender interface and Target/Response types
+│   │   ├── session.go            # SessionContext: thread-safe {{varName}} interpolation
+│   │   ├── session_test.go
 │   │   ├── warmup_executor.go    # WarmupExecutor: dispatches to HTTP or gRPC sender
 │   │   └── warmup_executor_test.go
 │   └── webhook/
@@ -249,6 +258,8 @@ kube-booster/
 │       ├── pod_mutator.go        # Webhook handler
 │       └── pod_mutator_test.go
 ├── config/
+│   ├── crd/                     # Custom Resource Definitions
+│   │   └── warmupconfig.yaml     # WarmupConfig CRD manifest
 │   ├── rbac/                    # RBAC manifests
 │   │   ├── service_account.yaml
 │   │   ├── role.yaml
@@ -261,11 +272,13 @@ kube-booster/
 │   │   └── daemonset.yaml        # Controller DaemonSet (node-local)
 │   ├── samples/                 # Sample applications
 │   │   ├── sample_deployment.yaml
-│   │   ├── sample_grpc_deployment.yaml  # gRPC warmup example
-│   │   └── grafana-dashboard.json  # Sample Grafana dashboard
+│   │   ├── sample_grpc_deployment.yaml     # gRPC warmup example
+│   │   ├── sample_warmup_config.yaml       # WarmupConfig CRD example
+│   │   └── sample_scenario_deployment.yaml # Deployment referencing WarmupConfig
 │   └── kustomization.yaml       # Kustomize config
 ├── hack/
 │   ├── generate_certs.sh        # Certificate generation
+│   ├── grpc-demo-server/        # Minimal gRPC demo server for testing
 │   └── quick_test.sh            # Smoke tests
 ├── Dockerfile                   # Multi-stage build
 ├── Makefile                     # Build automation
@@ -440,9 +453,11 @@ In this architecture, both webhook and controller run in a single deployment:
 - Context-aware cancellation support
 
 **http_sender.go**
-- `HTTPSender` sends a single HTTP GET request
+- `HTTPSender` sends a single HTTP request (any method with optional body)
+- Method defaults to `GET` when `Target.Method` is empty
+- `Target.Payload` bytes are sent as the request body when non-empty
 - Per-request timeout: 10s via `http.Client.Timeout`
-- Custom headers: `User-Agent: kube-booster/1.0`, `X-Warmup-Request: true`
+- Response body is captured in `Response.Body` for scenario response chaining
 
 **grpc_sender.go**
 - `GRPCSender` sends a single gRPC unary call using dynamic proto reflection
@@ -462,10 +477,26 @@ In this architecture, both webhook and controller run in a single deployment:
   - `kube-booster.io/warmup-port` → Port (auto-detected if possible)
   - `kube-booster.io/warmup-grpc-method` → gRPC method (`package.Service/Method`), required for gRPC
   - `kube-booster.io/warmup-grpc-payload` → JSON payload for gRPC request (default: `{}`)
+- `kube-booster.io/warmup-config` → Name of a `WarmupConfig` CR (enables scenario executor)
 - Validates `warmup-grpc-method` format and `warmup-grpc-payload` JSON validity at parse time
 - Auto-detects port from container spec (single container, single port)
 - `BuildEndpointURL()` constructs full URL for HTTP requests
 - `BuildGRPCAddress()` constructs `host:port` for gRPC dial
+
+**scenario_executor.go**
+- `ScenarioExecutorIface` interface: `ExecuteScenario(ctx, config, spec) *Result`
+- `ScenarioExecutor` orchestrates multi-step warmup defined in a `WarmupConfig` CR
+- Steps execute sequentially; within a step, requests execute in order
+- Per-request `{{varName}}` interpolation via `SessionContext`
+- JSON response extraction: simple dot-path only (`$.key`, `$.a.b`; no arrays or filters)
+- Per-step and overall context timeouts; step timeout expiry is fail-open (next step continues)
+- Reuses `HTTPSender` (arbitrary method + body) and `GRPCSender` (new per step for method isolation)
+- Rate-limited via shared `RequestRateLimiter` (same pool as `WarmupExecutor`)
+
+**session.go**
+- `SessionContext` is a thread-safe `map[string]any` with `Set`, `Get`, and `Interpolate` methods
+- `Interpolate(s)` replaces `{{varName}}` tokens; unknown keys are left unchanged
+- Safe for concurrent use via `sync.RWMutex` (future-proofed for parallel step execution)
 
 **result.go**
 - `Result` struct tracks warmup outcome:
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -274,7 +274,8 @@ kube-booster/
 │   │   ├── sample_deployment.yaml
 │   │   ├── sample_grpc_deployment.yaml     # gRPC warmup example
 │   │   ├── sample_warmup_config.yaml       # WarmupConfig CRD example
-│   │   └── sample_scenario_deployment.yaml # Deployment referencing WarmupConfig
+│   │   ├── sample_scenario_deployment.yaml # Deployment referencing WarmupConfig
+│   │   └── grafana-dashboard.json          # Sample Grafana dashboard
 │   └── kustomization.yaml       # Kustomize config
 ├── hack/
 │   ├── generate_certs.sh        # Certificate generation
@@ -458,6 +459,7 @@ In this architecture, both webhook and controller run in a single deployment:
 - `Target.Payload` bytes are sent as the request body when non-empty
 - Per-request timeout: 10s via `http.Client.Timeout`
 - Response body is captured in `Response.Body` for scenario response chaining
+- Custom headers: `User-Agent: kube-booster/1.0`, `X-Warmup-Request: true`
 
 **grpc_sender.go**
 - `GRPCSender` sends a single gRPC unary call using dynamic proto reflection
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -143,6 +143,7 @@ spec:
 | `kube-booster.io/warmup-port` | Container port for warmup requests | Auto-detected |
 | `kube-booster.io/warmup-grpc-method` | Fully-qualified gRPC method (`package.Service/Method`). Required when `warmup-protocol` is `grpc` | — |
 | `kube-booster.io/warmup-grpc-payload` | JSON-encoded request payload for gRPC warmup | `{}` |
+| `kube-booster.io/warmup-config` | Name of a `WarmupConfig` CR in the same namespace. When set, uses scenario-based warmup instead of single-endpoint warmup. See [WarmupConfig CRD](#warmupconfig-crd) | — |
 
 ### Example: Complete Application
 
@@ -251,6 +252,90 @@ See [`config/samples/sample_grpc_deployment.yaml`](../config/samples/sample_grpc
 - **Server reflection required**: The gRPC server must enable [server reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md) so kube-booster can discover method descriptors at warmup time without compiled proto files. In Go: `reflection.Register(grpcServer)`. In Java: `ProtoReflectionService`. In Python: `from grpc_reflection.v1alpha import reflection`.
 - **Unary RPCs only**: Only unary (non-streaming) RPCs are supported. Client-streaming, server-streaming, and bidirectional-streaming methods are rejected with a `WarmupFailed` event. Use a unary RPC such as a health-check or a lightweight read-only call.
 - **Plaintext transport**: gRPC warmup connections use plaintext (`insecure.NewCredentials()`). All warmup traffic between the controller and pod is unencrypted. Pod-to-pod traffic within a cluster is commonly treated as trusted, but if your security policy requires in-cluster encryption, apply a NetworkPolicy restricting controller-to-pod traffic on the warmup port while optional TLS support is tracked separately.
+
+### WarmupConfig CRD
+
+For applications that need multi-step warmup (e.g. load a cache, prime a recommendation engine, then verify health), use the `WarmupConfig` custom resource. Steps are executed sequentially; within a step, requests are executed in order.
+
+**Key features:**
+- **Multi-step warmup** with per-step and overall timeouts
+- **Response chaining**: extract JSON values from one response and inject them into the next request via `{{varName}}` interpolation
+- **Arbitrary HTTP methods** (GET, POST, PUT, etc.) with request bodies
+- **gRPC steps** mixed with HTTP steps in the same scenario
+- **Repeat count** per request to hit JIT compile thresholds
+
+**Usage:**
+
+1. Create a `WarmupConfig` CR in the same namespace as your pods:
+
+```yaml
+apiVersion: kube-booster.io/v1alpha1
+kind: WarmupConfig
+metadata:
+  name: my-app-warmup
+  namespace: default
+spec:
+  timeout: "120s"   # overall budget across all steps
+  steps:
+    - name: load-cache
+      timeout: "30s"
+      requests:
+        - name: preload
+          endpoint: /api/cache/load
+          method: POST
+          headers:
+            Content-Type: application/json
+          body: '{"action":"preload"}'
+          # Extract session.token from the JSON response body.
+          # Supported JSONPath syntax: $.key and $.a.b (no arrays or filters).
+          extract:
+            token: "$.session.token"
+
+    - name: prime-recommendations
+      requests:
+        - endpoint: /api/recs/prime
+          method: POST
+          headers:
+            Authorization: "Bearer {{token}}"  # interpolated from step 1
+          body: '{"userId":"warmup"}'
+
+    - name: health-check
+      requests:
+        - endpoint: /health
+          expectedStatus: 200
+          count: 3   # send 3 requests to trigger JIT compilation
+```
+
+2. Reference the `WarmupConfig` from your pod template:
+
+```yaml
+annotations:
+  kube-booster.io/warmup: "enabled"
+  kube-booster.io/warmup-config: "my-app-warmup"
+  kube-booster.io/warmup-port: "8080"    # still required
+```
+
+See [`config/samples/sample_warmup_config.yaml`](../config/samples/sample_warmup_config.yaml) and [`config/samples/sample_scenario_deployment.yaml`](../config/samples/sample_scenario_deployment.yaml) for complete examples.
+
+**`WarmupRequest` fields:**
+
+| Field | Description | Default |
+|-------|-------------|---------|
+| `name` | Optional label for log output | — |
+| `protocol` | `http` or `grpc` | inherited from pod annotation or `http` |
+| `endpoint` | URL path for HTTP requests | `/` |
+| `method` | HTTP verb | `GET` |
+| `headers` | HTTP request headers; supports `{{varName}}` | — |
+| `body` | HTTP request body; supports `{{varName}}` | — |
+| `grpcMethod` | Fully-qualified gRPC method (`pkg.Service/Method`) | — |
+| `grpcPayload` | JSON gRPC request message; supports `{{varName}}` | `{}` |
+| `count` | Number of times to repeat this request | `1` |
+| `extract` | `varName → $.json.path` mapping; extracted from last response body | — |
+| `expectedStatus` | HTTP status code that counts as success; `0` means 200–399 | `0` |
+
+**JSONPath extraction limitations:** Only simple dot-paths are supported (`$.key`, `$.a.b`). Array indexing and filter expressions are not supported.
+
+**Fail-open behavior:** If a step times out or a request fails, the scenario continues. The final result is fail-open just like single-endpoint warmup — the pod is marked READY regardless. If the `WarmupConfig` CR is not found, the controller falls back to annotation-based warmup with a `WarmupFailed` warning event.
 
 ### Controller Flags
 
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -262,7 +262,7 @@ For applications that need multi-step warmup (e.g. load a cache, prime a recomme
 - **Response chaining**: extract JSON values from one response and inject them into the next request via `{{varName}}` interpolation
 - **Arbitrary HTTP methods** (GET, POST, PUT, etc.) with request bodies
 - **gRPC steps** mixed with HTTP steps in the same scenario
-- **Repeat count** per request to hit JIT compile thresholds
+- **Repeat count** per request to warm up caches or trigger runtime optimization thresholds
 
 **Usage:**
 
@@ -296,6 +296,8 @@ spec:
         - endpoint: /api/recs/prime
           method: POST
           headers:
+            # NOTE: in production, use a dedicated warmup service account or
+            # short-lived credential rather than hardcoding tokens here.
             Authorization: "Bearer {{token}}"  # interpolated from step 1
           body: '{"userId":"warmup"}'
 
@@ -303,7 +305,7 @@ spec:
       requests:
         - endpoint: /health
           expectedStatus: 200
-          count: 3   # send 3 requests to trigger JIT compilation
+          count: 3   # send 3 requests to warm up caches
 ```
 
 2. Reference the `WarmupConfig` from your pod template:
@@ -335,7 +337,7 @@ See [`config/samples/sample_warmup_config.yaml`](../config/samples/sample_warmup
 
 **JSONPath extraction limitations:** Only simple dot-paths are supported (`$.key`, `$.a.b`). Array indexing and filter expressions are not supported.
 
-**Fail-open behavior:** If a step times out or a request fails, the scenario continues. The final result is fail-open just like single-endpoint warmup — the pod is marked READY regardless. If the `WarmupConfig` CR is not found, the controller falls back to annotation-based warmup with a `WarmupFailed` warning event.
+**Fail-open behavior:** If a step times out or a request fails, the scenario continues. The final result is fail-open just like single-endpoint warmup — the pod is marked READY regardless. If the `WarmupConfig` CR is not found, the controller emits a `WarmupFailed` warning event and fails open (pod is still marked READY); it does not fall back to annotation-based warmup.
 
 ### Controller Flags
 
--- a/pkg/api/v1alpha1/deepcopy.go
+++ b/pkg/api/v1alpha1/deepcopy.go
@@ -1,0 +1,98 @@
+package v1alpha1
+
+import "k8s.io/apimachinery/pkg/runtime"
+
+// DeepCopyObject implements runtime.Object.
+func (in *WarmupConfig) DeepCopyObject() runtime.Object {
+	if in == nil {
+		return nil
+	}
+	out := new(WarmupConfig)
+	in.DeepCopyInto(out)
+	return out
+}
+
+// DeepCopyInto copies all properties into another WarmupConfig.
+func (in *WarmupConfig) DeepCopyInto(out *WarmupConfig) {
+	*out = *in
+	out.TypeMeta = in.TypeMeta
+	in.ObjectMeta.DeepCopyInto(&out.ObjectMeta)
+	in.Spec.DeepCopyInto(&out.Spec)
+}
+
+// DeepCopy returns a deep copy of WarmupConfig.
+func (in *WarmupConfig) DeepCopy() *WarmupConfig {
+	if in == nil {
+		return nil
+	}
+	out := new(WarmupConfig)
+	in.DeepCopyInto(out)
+	return out
+}
+
+// DeepCopyObject implements runtime.Object.
+func (in *WarmupConfigList) DeepCopyObject() runtime.Object {
+	if in == nil {
+		return nil
+	}
+	out := new(WarmupConfigList)
+	in.DeepCopyInto(out)
+	return out
+}
+
+// DeepCopyInto copies all properties into another WarmupConfigList.
+func (in *WarmupConfigList) DeepCopyInto(out *WarmupConfigList) {
+	*out = *in
+	out.TypeMeta = in.TypeMeta
+	in.ListMeta.DeepCopyInto(&out.ListMeta)
+	if in.Items != nil {
+		in, out := &in.Items, &out.Items
+		*out = make([]WarmupConfig, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
+}
+
+// DeepCopyInto copies all properties into another WarmupConfigSpec.
+func (in *WarmupConfigSpec) DeepCopyInto(out *WarmupConfigSpec) {
+	*out = *in
+	if in.Steps != nil {
+		in, out := &in.Steps, &out.Steps
+		*out = make([]WarmupStep, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
+}
+
+// DeepCopyInto copies all properties into another WarmupStep.
+func (in *WarmupStep) DeepCopyInto(out *WarmupStep) {
+	*out = *in
+	if in.Requests != nil {
+		in, out := &in.Requests, &out.Requests
+		*out = make([]WarmupRequest, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
+}
+
+// DeepCopyInto copies all properties into another WarmupRequest.
+func (in *WarmupRequest) DeepCopyInto(out *WarmupRequest) {
+	*out = *in
+	if in.Headers != nil {
+		in, out := &in.Headers, &out.Headers
+		*out = make(map[string]string, len(*in))
+		for k, v := range *in {
+			(*out)[k] = v
+		}
+	}
+	if in.Extract != nil {
+		in, out := &in.Extract, &out.Extract
+		*out = make(map[string]string, len(*in))
+		for k, v := range *in {
+			(*out)[k] = v
+		}
+	}
+}
--- a/pkg/api/v1alpha1/register.go
+++ b/pkg/api/v1alpha1/register.go
@@ -1,0 +1,25 @@
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// SchemeGroupVersion is the group/version for kube-booster CRDs.
+var SchemeGroupVersion = schema.GroupVersion{Group: "kube-booster.io", Version: "v1alpha1"}
+
+// SchemeBuilder registers the v1alpha1 types.
+var (
+	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+	AddToScheme   = SchemeBuilder.AddToScheme
+)
+
+func addKnownTypes(scheme *runtime.Scheme) error {
+	scheme.AddKnownTypes(SchemeGroupVersion, &WarmupConfig{}, &WarmupConfigList{})
+	return nil
+}
+
+// Resource returns a GroupResource for the given resource name.
+func Resource(resource string) schema.GroupResource {
+	return SchemeGroupVersion.WithResource(resource).GroupResource()
+}
--- a/pkg/api/v1alpha1/types.go
+++ b/pkg/api/v1alpha1/types.go
@@ -1,0 +1,117 @@
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// WarmupConfig defines multi-step, scenario-based warmup requests for a pod.
+// A pod selects a WarmupConfig by setting the kube-booster.io/warmup-config
+// annotation to the name of the WarmupConfig in the same namespace.
+type WarmupConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec WarmupConfigSpec `json:"spec"`
+}
+
+// WarmupConfigList contains a list of WarmupConfig.
+type WarmupConfigList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []WarmupConfig `json:"items"`
+}
+
+// WarmupConfigSpec is the desired state of a WarmupConfig.
+type WarmupConfigSpec struct {
+	// Steps is the ordered list of warmup steps. Steps are executed sequentially.
+	// +kubebuilder:validation:MinItems=1
+	Steps []WarmupStep `json:"steps"`
+
+	// Timeout is the overall time limit for all steps combined.
+	// Parsed as a Go duration string (e.g. "120s", "2m"). Default: "120s".
+	// +optional
+	Timeout string `json:"timeout,omitempty"`
+}
+
+// WarmupStep groups one or more requests that are executed as a unit.
+type WarmupStep struct {
+	// Name is an optional human-readable label for the step (used in logs and events).
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// Requests is the list of warmup requests executed within this step. Requests
+	// are executed sequentially in the order they appear.
+	// +kubebuilder:validation:MinItems=1
+	Requests []WarmupRequest `json:"requests"`
+
+	// Timeout is the time limit for this step. Parsed as a Go duration string.
+	// Default: "30s".
+	// +optional
+	Timeout string `json:"timeout,omitempty"`
+}
+
+// WarmupRequest describes a single warmup call within a step.
+type WarmupRequest struct {
+	// Name is an optional label used in log output.
+	// +optional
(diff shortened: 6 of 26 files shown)


pkg/controller/pod_controller.go
```diff
@@ -151,10 +154,26 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
+		if err := r.Get(ctx, types.NamespacedName{
+			Name:      config.WarmupConfigName,
+			Namespace: pod.Namespace,
+		}, warmupCfg); err != nil {
+			// WarmupConfig not found: log a warning and fall back to annotation-based warmup.
+			logger.Error(err, "WarmupConfig not found, falling back to annotation-based warmup",
+				"warmupConfig", config.WarmupConfigName)
+			r.Recorder.Eventf(pod, nil, corev1.EventTypeWarning, ReasonWarmupFailed, "LookupWarmupConfig",
+				"WarmupConfig %q not found: %v (fail-open, using annotation-based warmup)", config.WarmupConfigName, err)
+			result = r.WarmupExecutor.Execute(warmupCtx, config)
```
hhiroshell: **[Logic]** :rotating_light: **Nil panic in WarmupExecutor fallback path.**

When the WarmupConfig CR is not found, the code falls back to `r.WarmupExecutor.Execute(...)` here without a nil check. The outer `else if r.WarmupExecutor != nil` guard does NOT protect this inner path. A `PodReconciler` constructed with `WarmupExecutor: nil` (and only `ScenarioExecutor` set) will panic here if it encounters a pod with a `warmup-config` annotation pointing to a non-existent CR.

Fix:
```go
if r.WarmupExecutor != nil {
    result = r.WarmupExecutor.Execute(warmupCtx, config)
} else {
    recordMetrics = false
    result = &warmup.Result{Success: true, Message: "warmup skipped: no executor configured"}
}
```

pkg/warmup/http_sender.go
```diff
@@ -34,14 +47,15 @@ func (s *HTTPSender) Send(ctx context.Context, target Target) *Response {
 		return &Response{Error: err, Duration: duration}
 	}
 
-	if _, drainErr := io.Copy(io.Discard, resp.Body); drainErr != nil {
-		s.logger.V(2).Info("failed to drain response body", "error", drainErr)
+	body, readErr := io.ReadAll(resp.Body)
```
hhiroshell: **[Security/Performance]** :warning: **Unbounded `io.ReadAll` allows memory exhaustion.**

There is no size limit on the response body read here. A pod returning a multi-gigabyte response can exhaust controller memory. This is amplified by scenario `count` repetitions (all responses are buffered).

Fix:
```go
const maxResponseBodyBytes = 1 << 20 // 1 MiB
body, readErr := io.ReadAll(io.LimitReader(resp.Body, maxResponseBodyBytes))
```
(CWE-400)

pkg/warmup/session.go
```diff
@@ -0,0 +1,52 @@
...
+// of the corresponding session value. Tokens referencing unknown keys are left as-is.
+// Safe for concurrent use.
+func (sc *SessionContext) Interpolate(s string) string {
+	if !strings.Contains(s, "{{") {
+		return s
+	}
+	sc.mu.RLock()
+	defer sc.mu.RUnlock()
+	for k, v := range sc.data {
+		s = strings.ReplaceAll(s, "{{"+k+"}}", fmt.Sprintf("%v", v))
```
hhiroshell: **[Security]** :warning: **Template injection via random map iteration order.**

Go map iteration is non-deterministic. If an extracted session value contains a `{{key}}` token — for example, a compromised app pod returns `{"token": "{{adminSecret}}"}` — and `adminSecret` happens to be iterated after `token` in this same `range` loop, the injected placeholder gets substituted. Application-controlled response data can thus influence what gets sent in subsequent request headers, bodies, or endpoint paths.

Fix: iterate keys in a sorted, deterministic order so already-substituted text is processed before the keys it might reference:
```go
keys := make([]string, 0, len(sc.data))
for k := range sc.data { keys = append(keys, k) }
sort.Strings(keys)
for _, k := range keys {
    s = strings.ReplaceAll(s, "{{"+k+"}}", fmt.Sprintf("%v", sc.data[k]))
}
```
(CWE-94)

pkg/warmup/scenario_executor.go
```diff
@@ -0,0 +1,306 @@
...
+	v1alpha1 "github.com/hhiroshell/kube-booster/pkg/api/v1alpha1"
+)
+
+const (
+	defaultScenarioTimeout = 120 * time.Second
+	defaultStepTimeout     = 30 * time.Second
+)
+
+// ScenarioExecutorIface is implemented by ScenarioExecutor and any test doubles.
+type ScenarioExecutorIface interface {
```
hhiroshell: **[Design/Convention]** :warning: **`ScenarioExecutorIface` naming violates project conventions.**

Every other interface in this package uses a clean noun without suffix: `Executor` (not `ExecutorIface`), `Sender` (not `SenderIface`). The `Iface` suffix is a Java-style anti-pattern inconsistent with idiomatic Go and the rest of this codebase.

Suggest renaming:
- Interface → `ScenarioExecutor`
- Concrete struct → `scenarioExecutor` (unexported) or keep as `ScenarioExecutor` with the interface renamed to `ScenarioExecutorI` — actually the cleanest Go pattern is: interface gets the short name, concrete type gets `Impl` suffix or stays unexported.

pkg/warmup/scenario_executor.go
```diff
@@ -0,0 +1,306 @@
...
+					interpolatedHeaders[k] = session.Interpolate(v)
+				}
+
+				method := req.Method
+				if method == "" {
+					method = http.MethodGet
+				}
+
+				resp = (&HTTPSender{client: e.httpClient, logger: e.logger}).Send(ctx, Target{
+					Address: fmt.Sprintf("http://%s:%d%s", config.PodIP, config.Port, endpoint),
```
hhiroshell: **[Design]** :warning: **Inline URL construction duplicates `Config.BuildEndpointURL()`.**

This block (leading-slash normalization + `fmt.Sprintf("http://%s:%d%s", ...)`) exactly replicates what `config.BuildEndpointURL()` already does in `config.go:208-215`. The gRPC path correctly uses `config.BuildGRPCAddress()` — HTTP should follow the same pattern.

`BuildEndpointURL` could be extended to accept an explicit endpoint path:
```go
func (c *Config) BuildEndpointURLFor(endpoint string) string {
    if len(endpoint) == 0 || endpoint[0] != '/' {
        endpoint = "/" + endpoint
    }
    return fmt.Sprintf("http://%s:%d%s", c.PodIP, c.Port, endpoint)
}
```
Then here:
```go
Address: config.BuildEndpointURLFor(endpoint),
```

pkg/warmup/scenario_executor.go
```diff
@@ -0,0 +1,306 @@
...
+				for k, v := range req.Headers {
+					interpolatedHeaders[k] = session.Interpolate(v)
+				}
+
+				method := req.Method
+				if method == "" {
+					method = http.MethodGet
+				}
+
+				resp = (&HTTPSender{client: e.httpClient, logger: e.logger}).Send(ctx, Target{
```
hhiroshell: **[Performance]** :warning: **`HTTPSender` allocated on every `count` iteration.**

`&HTTPSender{client: e.httpClient, logger: e.logger}` is a thin wrapper over the shared `e.httpClient` — it carries no per-request state. Allocating it inside the inner `count` loop creates unnecessary GC pressure for high-count scenarios. The `grpcSender` is correctly created once per step; `HTTPSender` should be created once per request entry (or once per step):
```go
httpSender := &HTTPSender{client: e.httpClient, logger: e.logger}
for i := 0; i < count; i++ {
    // ...
    resp = httpSender.Send(ctx, Target{...})
}
```

pkg/api/v1alpha1/deepcopy.go
```diff
@@ -0,0 +1,98 @@
+package v1alpha1
```
hhiroshell, line 1: **[Convention]** :warning: **No tests for hand-written `DeepCopy` methods.**

CLAUDE.md requires >80% coverage for new code. These three `DeepCopyInto`/`DeepCopy` implementations have zero test coverage. Incorrect deep copies (e.g. a missed slice re-allocation means modifying a copy mutates the original) are silent bugs that are easy to introduce and hard to diagnose.

Suggest adding `deepcopy_test.go` with mutation tests:
```go
func TestWarmupConfig_DeepCopy(t *testing.T) {
    orig := &WarmupConfig{Spec: WarmupConfigSpec{Steps: []WarmupStep{{Name: "s1"}}}}
    copy := orig.DeepCopy()
    copy.Spec.Steps[0].Name = "mutated"
    if orig.Spec.Steps[0].Name != "s1" {
        t.Error("DeepCopy shared slice with original")
    }
}
```

config/crd/warmupconfig.yaml
```diff
@@ -0,0 +1,94 @@
...
+                              minimum: 1
+                              description: "Number of times to repeat this request. Default: 1."
+                            extract:
+                              type: object
+                              additionalProperties:
+                                type: string
+                              description: "Map from session variable name to JSONPath expression ($.key or $.a.b). Extracted from last response body."
+                            expectedStatus:
+                              type: integer
+                              description: "HTTP status code that counts as success. When 0, 200–399 are success. Ignored for gRPC."
```
hhiroshell, line 108: **[Security]** :warning: **No `maxLength` constraints on string fields; no `maximum` on `count`.**

Two issues:

1. String fields (`endpoint`, `body`, `grpcPayload`, `grpcMethod`, `name`) have no `maxLength`. A user with write access to WarmupConfig resources can store multi-megabyte strings that are loaded into controller memory and forwarded on every reconcile. Add e.g. `maxLength: 65536` for `body`/`grpcPayload` and `maxLength: 1024` for path/method fields.

2. `count` has `minimum: 1` but no `maximum`. A value of `2147483647` causes the executor to loop for the full step timeout, burning CPU and rate-limiter tokens. The annotation-based path caps at `MaxRequestCount = 12000` (config.go:19); the CRD should enforce the same:
```yaml
count:
  type: integer
  minimum: 1
  maximum: 12000
```
(CWE-400)

pkg/warmup/scenario_executor.go
```diff
@@ -0,0 +1,306 @@
...
+	if protocol == ProtocolGRPC {
+		return statusCode == 200
+	}
+	if expectedStatus != 0 {
+		return statusCode == expectedStatus
+	}
+	return statusCode >= 200 && statusCode < 400
+}
+
+// extractVariables parses body as JSON and stores matched JSONPath values in session.
```
hhiroshell, line 270: **[Design]** :information_source: **`extractVariables` is a receiver method but uses no receiver state.**

`extractVariables` accesses only its parameters (body, extract map, session, reqName) and logs via `reqName` — it never reads `e.logger` or any other `ScenarioExecutor` field. `jsonPathLookup` defined just below is correctly a package-level function. For consistency and independent testability, `extractVariables` should also be package-level:
```go
func extractVariables(body []byte, extract map[string]string, session *SessionContext, logger logr.Logger, reqName string) {
```

pkg/controller/pod_controller.go
```diff
@@ -151,10 +154,26 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 	defer cancel()
 
 	var result *warmup.Result
-	if r.WarmupExecutor != nil {
+	recordMetrics := true
+	if config.WarmupConfigName != "" && r.ScenarioExecutor != nil {
+		warmupCfg := &v1alpha1.WarmupConfig{}
+		if err := r.Get(ctx, types.NamespacedName{
```
hhiroshell: **[Logic]** :information_source: **WarmupConfig CR lookup bypasses the warmup timeout.**

`r.Get(ctx, ...)` uses the outer reconcile context, not `warmupCtx` (which carries the `config.Timeout + 5s` deadline). If the API server is slow, this lookup can run well past the configured warmup timeout before `ExecuteScenario` even starts. Using `warmupCtx` here would keep the full warmup phase—CRD resolution included—within the configured time budget.

```go
if err := r.Get(warmupCtx, types.NamespacedName{...}, warmupCfg); err != nil {
```

hhiroshell (commented):

This review was generated by analyzing PR #70 from 8 perspectives.

### Summary

| Perspective | Findings |
|-------------|----------|
| Logic & Correctness | 3 issues |
| Design & Maintainability | 4 issues |
| Security | 3 issues |
| Performance | 2 issues |
| Convention Compliance | 2 issues |
| Git History Context | 0 issues |
| PR History Context | 0 issues |
| Documentation | 0 issues |
| **Total** | **14 issues** |

### Key Findings

#### :rotating_light: Errors (must fix)

**[Logic] Nil panic in WarmupExecutor fallback path** (`pkg/controller/pod_controller.go:169`)

When `WarmupConfigName` is set and `ScenarioExecutor != nil`, but the WarmupConfig CR is not found, the code falls back to `r.WarmupExecutor.Execute(...)` at line 169 without a nil check. The nil guard (`else if r.WarmupExecutor != nil`) only protects the outer branch, not this inner fallback. A `PodReconciler` constructed with only `ScenarioExecutor` set (and `WarmupExecutor = nil`) will panic when it encounters a missing WarmupConfig. Fix: add a nil guard to the fallback path, matching the pattern in the outer `else if`.

#### :warning: Warnings (should fix)

**[Security] Unbounded `io.ReadAll` on response body** (`pkg/warmup/http_sender.go:50`)

`io.ReadAll(resp.Body)` has no size limit. A pod (or compromised pod) returning a multi-gigabyte response body can exhaust controller memory. This is amplified by scenario `count` repetitions. Fix: wrap with `io.LimitReader(resp.Body, maxBodyBytes)` (e.g. 1 MiB cap). (CWE-400)

**[Security] Template injection via random map iteration in `SessionContext.Interpolate`** (`pkg/warmup/session.go:48`)

Go map iteration order is random. If an extracted session value contains a `{{key}}` token (e.g. a compromised pod returns `{"token": "{{adminSecret}}"}`) and `adminSecret` happens to be iterated after `token` in the same `range sc.data` loop, the injected token gets substituted into the final string. Application-controlled response data can thus influence subsequent request headers, body, or endpoint. Fix: escape the substituted value before continuing (e.g. perform a single ordered pass over sorted keys so already-substituted text is not re-scanned). (CWE-94/CWE-116)

**[Design] `ScenarioExecutorIface` naming violates project interface conventions** (`pkg/warmup/scenario_executor.go:22`)

The established pattern in this package is clean noun for the interface (`Executor`, `Sender`) and a descriptive concrete type (`WarmupExecutor`, `HTTPSender`, `GRPCSender`). The `Iface` suffix is a Java-style anti-pattern not used anywhere else in this codebase. The interface should be renamed to `ScenarioExecutor` and the concrete struct to something like `defaultScenarioExecutor` or `scenarioExecutor`.

**[Design] Inline URL construction duplicates `Config.BuildEndpointURL()`** (`pkg/warmup/scenario_executor.go:224`)

`fmt.Sprintf("http://%s:%d%s", config.PodIP, config.Port, endpoint)` replicates exactly what `config.BuildEndpointURL()` (config.go:208) already does, including the leading-slash normalization that is done redundantly at lines 202-208. The gRPC path already correctly uses `config.BuildGRPCAddress()`. `BuildEndpointURL` should be extended to accept an optional endpoint override, or a new helper `BuildEndpointURLFor(endpoint string)` added, to eliminate this duplication.

**[Performance/Security] `&HTTPSender{...}` allocated on every loop iteration** (`pkg/warmup/scenario_executor.go:223`)

A new `HTTPSender` value (and its struct allocation) is created for every iteration of the `count` inner loop, even though `HTTPSender` is a thin wrapper over the shared `e.httpClient`. The `GRPCSender` is correctly created once per step and cached. `HTTPSender` should be created once per request entry (outside the `count` loop) to match this pattern and eliminate GC pressure in scenarios with high `count` values.

**[Convention] No test coverage for hand-written `DeepCopy` methods** (`pkg/api/v1alpha1/deepcopy.go`)

CLAUDE.md requires >80% coverage for new code. The three hand-written `DeepCopyInto`/`DeepCopy` implementations for `WarmupConfig`, `WarmupConfigSpec`, and `WarmupStep`/`WarmupRequest` have no tests. Incorrect deep copies (e.g. missed slice re-allocation) are silent and easy to introduce. A `deepcopy_test.go` with mutation tests (modify the copy, assert the original is unchanged) would catch this.

**[Security] No `maxLength` constraints on CRD string fields** (`config/crd/warmupconfig.yaml`)

String fields `endpoint`, `body`, `grpcPayload`, `grpcMethod`, `name`, and header/extract values have no `maxLength` in the OpenAPI schema. A user with write access to WarmupConfig resources can store arbitrarily large strings that are loaded into memory and forwarded by the controller on every reconcile. Add `maxLength` to all unbounded string fields and `maxProperties` to the `headers` and `extract` maps. (CWE-400)

#### :information_source: Info (nice to fix)

**[Design] `extractVariables` is a receiver method but uses no receiver state** (`pkg/warmup/scenario_executor.go:268`)

`extractVariables` accesses only its parameters and the logger passed via `reqName`. `jsonPathLookup` (defined just below) is correctly a package-level function. Both should be package-level helpers for consistency and independent testability.

**[Logic] `r.Get` for WarmupConfig uses outer `ctx`, not `warmupCtx`** (`pkg/controller/pod_controller.go:160`)

The WarmupConfig lookup bypasses the warmup timeout boundary. If the API server is slow, this call can run well beyond the configured warmup timeout before `ExecuteScenario` even starts. Using `warmupCtx` here would keep the entire warmup phase—including CRD resolution—within the configured time budget.

**[Security] `count` field has no `maximum` in CRD schema** (`config/crd/warmupconfig.yaml:84`)

`count` has `minimum: 1` but no `maximum`. A value of `2147483647` causes the executor to loop for the entire step/scenario timeout period, consuming CPU and rate-limiter tokens continuously. The annotation-based path caps at `MaxRequestCount = 12000`; the CRD schema should enforce a consistent cap via `maximum: 12000`. (CWE-400)

---

docs/USAGE.md
```diff
@@ -252,6 +253,90 @@ See [`config/samples/sample_grpc_deployment.yaml`](../config/samples/sample_grpc
...
+| `method` | HTTP verb | `GET` |
+| `headers` | HTTP request headers; supports `{{varName}}` | — |
+| `body` | HTTP request body; supports `{{varName}}` | — |
+| `grpcMethod` | Fully-qualified gRPC method (`pkg.Service/Method`) | — |
+| `grpcPayload` | JSON gRPC request message; supports `{{varName}}` | `{}` |
+| `count` | Number of times to repeat this request | `1` |
+| `extract` | `varName → $.json.path` mapping; extracted from last response body | — |
+| `expectedStatus` | HTTP status code that counts as success; `0` means 200–399 | `0` |
+
+**JSONPath extraction limitations:** Only simple dot-paths are supported (`$.key`, `$.a.b`). Array indexing and filter expressions are not supported.
```
hhiroshell, line 338: What do you think about this limitation? Is it enough if we don't support array indexing and filter expressions?

docs/DEVELOPMENT.md
```diff
@@ -261,11 +272,13 @@ kube-booster/
 │   │   └── daemonset.yaml        # Controller DaemonSet (node-local)
 │   ├── samples/                 # Sample applications
 │   │   ├── sample_deployment.yaml
-│   │   ├── sample_grpc_deployment.yaml  # gRPC warmup example
-│   │   └── grafana-dashboard.json  # Sample Grafana dashboard
```
hhiroshell, line 265: Have you deleted the dashboard sample?

docs/DEVELOPMENT.md
```diff
@@ -440,9 +453,11 @@ In this architecture, both webhook and controller run in a single deployment:
 - Context-aware cancellation support
 
 **http_sender.go**
-- `HTTPSender` sends a single HTTP GET request
+- `HTTPSender` sends a single HTTP request (any method with optional body)
+- Method defaults to `GET` when `Target.Method` is empty
+- `Target.Payload` bytes are sent as the request body when non-empty
 - Per-request timeout: 10s via `http.Client.Timeout`
-- Custom headers: `User-Agent: kube-booster/1.0`, `X-Warmup-Request: true`
```
hhiroshell, line 445: Keep this custom header.

docs/DEVELOPMENT.md
```diff
@@ -462,11 +477,27 @@ In this architecture, both webhook and controller run in a single deployment:
...
+- Per-request `{{varName}}` interpolation via `SessionContext`
+- JSON response extraction: simple dot-path only (`$.key`, `$.a.b`; no arrays or filters)
+- Per-step and overall context timeouts; step timeout expiry is fail-open (next step continues)
+- Reuses `HTTPSender` (arbitrary method + body) and `GRPCSender` (new per step for method isolation)
+- Rate-limited via shared `RequestRateLimiter` (same pool as `WarmupExecutor`)
+
+**session.go**
+- `SessionContext` is a thread-safe `map[string]any` with `Set`, `Get`, and `Interpolate` methods
+- `Interpolate(s)` replaces `{{varName}}` tokens; unknown keys are left unchanged
+- Safe for concurrent use via `sync.RWMutex` (future-proofed for parallel step execution)
```
hhiroshell, line 501: 👍

docs/USAGE.md
```diff
@@ -252,6 +253,90 @@ See [`config/samples/sample_grpc_deployment.yaml`](../config/samples/sample_grpc
 - **Unary RPCs only**: Only unary (non-streaming) RPCs are supported. Client-streaming, server-streaming, and bidirectional-streaming methods are rejected with a `WarmupFailed` event. Use a unary RPC such as a health-check or a lightweight read-only call.
 - **Plaintext transport**: gRPC warmup connections use plaintext (`insecure.NewCredentials()`). All warmup traffic between the controller and pod is unencrypted. Pod-to-pod traffic within a cluster is commonly treated as trusted, but if your security policy requires in-cluster encryption, apply a NetworkPolicy restricting controller-to-pod traffic on the warmup port while optional TLS support is tracked separately.
 
+### WarmupConfig CRD
+
+For applications that need multi-step warmup (e.g. load a cache, prime a recommendation engine, then verify health), use the `WarmupConfig` custom resource. Steps are executed sequentially; within a step, requests are executed in order.
+
+**Key features:**
+- **Multi-step warmup** with per-step and overall timeouts
+- **Response chaining**: extract JSON values from one response and inject them into the next request via `{{varName}}` interpolation
```
hhiroshell, line 262: At the current phase, I tihnk it's enough to support only JSON responses. However what do you think about supporting other formats in the future?

docs/USAGE.md
```diff
@@ -252,6 +253,90 @@ See [`config/samples/sample_grpc_deployment.yaml`](../config/samples/sample_grpc
...
+- **Arbitrary HTTP methods** (GET, POST, PUT, etc.) with request bodies
+- **gRPC steps** mixed with HTTP steps in the same scenario
+- **Repeat count** per request to hit JIT compile thresholds
```
hhiroshell: Actually, we set the request limit to cover JIT compilation use cases, but it's not suitable to single out JIT specifically here.

docs/USAGE.md
```diff
@@ -252,6 +253,90 @@ See [`config/samples/sample_grpc_deployment.yaml`](../config/samples/sample_grpc
...
+        - name: preload
+          endpoint: /api/cache/load
+          method: POST
+          headers:
+            Content-Type: application/json
+          body: '{"action":"preload"}'
+          # Extract session.token from the JSON response body.
+          # Supported JSONPath syntax: $.key and $.a.b (no arrays or filters).
+          extract:
+            token: "$.session.token"
```
hhiroshell, line 292: It returns a session token in response to simple requests without any identity. Is this usual behavior for general web applications or APIs?

docs/USAGE.md
```diff
@@ -252,6 +253,90 @@ See [`config/samples/sample_grpc_deployment.yaml`](../config/samples/sample_grpc
...
+| `body` | HTTP request body; supports `{{varName}}` | — |
+| `grpcMethod` | Fully-qualified gRPC method (`pkg.Service/Method`) | — |
+| `grpcPayload` | JSON gRPC request message; supports `{{varName}}` | `{}` |
+| `count` | Number of times to repeat this request | `1` |
+| `extract` | `varName → $.json.path` mapping; extracted from last response body | — |
+| `expectedStatus` | HTTP status code that counts as success; `0` means 200–399 | `0` |
+
+**JSONPath extraction limitations:** Only simple dot-paths are supported (`$.key`, `$.a.b`). Array indexing and filter expressions are not supported.
+
+**Fail-open behavior:** If a step times out or a request fails, the scenario continues. The final result is fail-open just like single-endpoint warmup — the pod is marked READY regardless. If the `WarmupConfig` CR is not found, the controller falls back to annotation-based warmup with a `WarmupFailed` warning event.
```
hhiroshell: I think this fallback behavior is not easy for users to expect. What do you think about simlply falling back to a warm-up failure and leading to a fail-open path?

pkg/api/v1alpha1/deepcopy.go
hhiroshell: These implementations should be generated by the Kubebuilder toolkit, right?

pkg/api/v1alpha1/register.go
```diff
@@ -0,0 +1,25 @@
+package v1alpha1
```
hhiroshell, line 1: Same as the comment on `deepcopy.go` (this might be generated rather than handwritten.)

pkg/api/v1alpha1/types.go
```diff
@@ -0,0 +1,117 @@
+package v1alpha1
```
hhiroshell, line 1: Same as the comment on `deepcopy.go` (this might be generated rather than handwritten.)

pkg/controller/pod_controller_test.go
```diff
@@ -619,6 +620,174 @@ func hasPrefix(s, prefix string) bool {
...
+				strings.Contains(ev, "does-not-exist") {
+				foundWarning = true
+			}
+		}
+		if !foundWarning {
+			t.Error("expected Warning/WarmupFailed event mentioning the missing WarmupConfig name")
+		}
+	})
+
+	t.Run("no WarmupConfigName uses WarmupExecutor as before", func(t *testing.T) {
```
hhiroshell: The phrase "as before" is confusing for newbie developers who don't know the context.

pkg/warmup/scenario_executor.go
```diff
@@ -0,0 +1,306 @@
...
+		result.Message = "cannot execute scenario: pod IP not set"
+		return result
+	}
+
+	overallTimeout := defaultScenarioTimeout
+	if spec.Timeout != "" {
+		if d, err := time.ParseDuration(spec.Timeout); err == nil && d > 0 {
+			overallTimeout = d
+		}
+	}
```
hhiroshell, line 83: Are there any checks for the upper limit of scenarioTimeout? Without validation, users could set excessively long timeouts and occupy the execution queue.

pkg/warmup/scenario_executor_test.go
```diff
@@ -0,0 +1,383 @@
...
+				Requests: []v1alpha1.WarmupRequest{
+					{Endpoint: "/quick"},
+				},
+			},
+		},
+	}
+
+	result := e.ExecuteScenario(context.Background(), config, spec)
+	// Step 1 timed out (counted as failed), step 2 should succeed
+	_ = step1Done
```
hhiroshell: It looks like this test never checks if step1 exits due to the step timeout.

pkg/controller/pod_controller.go
hhiroshell: Confirmed bug. Will add a nil guard around the fallback:

```go
if r.WarmupExecutor != nil {
    result = r.WarmupExecutor.Execute(warmupCtx, config)
} else {
    recordMetrics = false
    result = &warmup.Result{Success: true, Message: "warmup skipped: no executor configured"}
}
```

Will also extend `TestPodReconciler_ScenarioDispatch` with a sub-case that constructs `PodReconciler` with `WarmupExecutor: nil` to catch this regression going forward.

---

pkg/warmup/http_sender.go
hhiroshell: Valid concern. Will add `io.LimitReader` with a 1 MiB cap:

```go
const maxResponseBodyBytes = 1 << 20 // 1 MiB
body, readErr := io.ReadAll(io.LimitReader(resp.Body, maxResponseBodyBytes))
```

If a response is larger than 1 MiB, extraction works on the truncated body — which is fine for typical token/ID extraction patterns. Will also expose the cap as a package-level constant so it can be tuned in the future.

---

pkg/warmup/session.go
hhiroshell: Good catch. Sorting keys before the loop eliminates the non-determinism:

```go
keys := make([]string, 0, len(sc.data))
for k := range sc.data { keys = append(keys, k) }
sort.Strings(keys)
for _, k := range keys {
    s = strings.ReplaceAll(s, "{{"+k+"}}", fmt.Sprintf("%v", sc.data[k]))
}
```

This ensures already-substituted text is never re-scanned for new placeholders, so application-controlled response values cannot inject `{{otherKey}}` tokens into subsequent substitutions. Will implement.

---

pkg/warmup/scenario_executor.go
hhiroshell: Agreed — the `Iface` suffix is out of place here. Will rename:

- Interface: `ScenarioExecutor` (matches the `Executor`/`Sender` pattern used everywhere else)
- Concrete struct: `defaultScenarioExecutor` (unexported — all callers go through the interface)
- Constructor: `NewScenarioExecutor` stays, returning the interface type

`MockScenarioExecutor` in `mock.go` stays as-is since it's a test double.

---

pkg/warmup/scenario_executor.go
hhiroshell: Agreed. Will add `BuildEndpointURLFor` to `Config` to remove the duplication:

```go
// BuildEndpointURLFor constructs the full URL for a given endpoint path,
// applying the same leading-slash normalization as BuildEndpointURL.
func (c *Config) BuildEndpointURLFor(endpoint string) string {
    if len(endpoint) == 0 || endpoint[0] != '/' {
        endpoint = "/" + endpoint
    }
    return fmt.Sprintf("http://%s:%d%s", c.PodIP, c.Port, endpoint)
}
```

The scenario executor can then replace the inline block with `config.BuildEndpointURLFor(endpoint)`, matching how the gRPC path uses `config.BuildGRPCAddress()`.

---

pkg/warmup/scenario_executor.go
hhiroshell: Agreed. Will move the `HTTPSender` construction above the `count` loop, matching the `grpcSender` pattern:

```go
httpSender := &HTTPSender{client: e.httpClient, logger: e.logger}
for i := 0; i < count; i++ {
    // ...
    resp = httpSender.Send(ctx, Target{...})
}
```

`HTTPSender` carries no per-request state, so reuse is safe.

---

pkg/api/v1alpha1/deepcopy.go
hhiroshell: Yes, under a typical Kubebuilder project these three files would be generated by `controller-gen` from `+k8s:deepcopy-gen` markers on the types. We hand-wrote them because the project doesn't have a code-gen step in the Makefile — the `pkg/api/v1alpha1/` package was added manually without integrating the controller-gen toolchain.

For now, will add the suggested mutation tests to `deepcopy_test.go` to catch correctness regressions. I'll open a separate issue to evaluate adding `controller-gen` so that `types.go`, `register.go`, and `deepcopy.go` can all be properly generated — that would also let us use `+kubebuilder:validation` markers for admission-time field validation instead of relying on runtime checks.

---

config/crd/warmupconfig.yaml
hhiroshell: Will add both constraints. Proposed values aligned with the annotation-based path:

```yaml
# string fields
endpoint / grpcMethod / name:  maxLength: 1024
body / grpcPayload:            maxLength: 65536
headers / extract values:      maxLength: 4096

# count
count:
  type: integer
  minimum: 1
  maximum: 12000   # matches MaxRequestCount in config.go
```

For `headers` and `extract` maps, will also add `maxProperties: 50` to bound the number of entries.

---

pkg/warmup/scenario_executor.go
hhiroshell: Agreed. Will convert to a package-level function:

```go
func extractVariables(body []byte, extract map[string]string, session *SessionContext, logger logr.Logger, reqName string) {
```

This matches `jsonPathLookup` just below, makes both functions independently testable without constructing a full executor, and makes the stateless nature explicit.

---

pkg/controller/pod_controller.go
hhiroshell: Good point. Will switch to `warmupCtx`:

```go
if err := r.Get(warmupCtx, types.NamespacedName{
    Name:      config.WarmupConfigName,
    Namespace: pod.Namespace,
}, warmupCfg); err != nil {
```

This keeps the entire warmup phase — including CRD resolution — within the configured timeout budget. The error path (fallback + Warning event) is unaffected.

---

docs/USAGE.md
hhiroshell: For the primary warmup use case — extracting a session token, an ID, or a config value from a structured response — simple dot-path (`$.key`, `$.a.b`) is sufficient. The most common patterns in warmup scenarios are single scalar values.

Full JSONPath (array indexing `$.items[0].id`, filter expressions `$.items[?(@.status=='active')]`) would require pulling in a library dependency (e.g. `github.com/PaesslerAG/jsonpath`) and makes the implementation harder to audit. I think the current scope is the right call for v1alpha1 — the important thing is that the limitation is clearly documented (which it is at line 336). If users report hitting this, we can evaluate adding a library at that point.

---

docs/DEVELOPMENT.md
hhiroshell: No — `config/samples/grafana-dashboard.json` still exists on disk. The PR accidentally dropped it from the directory tree listing in DEVELOPMENT.md when restructuring that block to add the new sample files (`sample_warmup_config.yaml`, `sample_scenario_deployment.yaml`). Will restore the entry:

```
│   │   ├── sample_warmup_config.yaml       # WarmupConfig CRD example
│   │   ├── sample_scenario_deployment.yaml # Deployment referencing WarmupConfig
│   │   └── grafana-dashboard.json          # Sample Grafana dashboard
```

---

docs/DEVELOPMENT.md
hhiroshell: You're right — the line `Custom headers: \`User-Agent: kube-booster/1.0\`, \`X-Warmup-Request: true\`` was unintentionally dropped when the HTTPSender description was updated to document the new method/body/response-body capabilities. The headers are still added in `scenario_executor.go` (lines 212–213), so the documentation just fell behind. Will restore it alongside the new lines.

---

docs/DEVELOPMENT.md
hhiroshell: 👍

---

docs/USAGE.md
hhiroshell: Agree JSON-only is the right scope for now. If we ever add other formats, the cleanest extension point would be a `format` field on `WarmupRequest` (`"json"`, `"text"`) with format-specific extraction dispatched in `extractVariables`. For gRPC, response extraction would need a protobuf field path rather than JSONPath. I'd keep all of that out of scope until a concrete user need arises.

---

docs/USAGE.md
hhiroshell: Agreed — "hit JIT compile thresholds" is too narrow. `count` is useful for any scenario that benefits from repeated calls: cache warming, connection pool preheating, code path pre-loading, etc. Will update the description to something like "send multiple requests to the same endpoint to warm up caches or trigger runtime optimization thresholds" without singling out JIT.

---

docs/USAGE.md
hhiroshell: Fair observation — the sample is simplified for clarity. In practice, internal cache-preload endpoints in a microservice architecture are typically either:

1. Protected at the network layer (cluster-internal only, via `NetworkPolicy`)
2. Using mTLS (service mesh) so no explicit application-level auth is needed
3. Accepting a static shared secret or service-account token in a header

The sample strips out auth to focus on the response-chaining concept. Will add a comment in the YAML noting that production usage would typically include authentication credentials in the request headers.

---

docs/USAGE.md
hhiroshell: Good point — the current double-fallback (WarmupConfig not found → silently use annotation-based warmup) violates the principle of least surprise. The user explicitly requested scenario-based warmup; getting a different warmup without notice is confusing.

The clearer behavior is: WarmupConfig not found → emit `Warning/WarmupFailed` event → return a failed `Result` → fail-open (pod becomes READY anyway). This is consistent with how all other warmup failures are handled. Will change the fallback path accordingly and update the USAGE.md documentation.

---

pkg/api/v1alpha1/register.go
hhiroshell: Same answer as for `deepcopy.go` — `register.go` would normally be generated by `controller-gen` from `+groupversion-info` markers. It's hand-written here because the project doesn't yet have a code-gen step. Will track adding `controller-gen` in a dedicated issue so all three files (`types.go`, `register.go`, `deepcopy.go`) can be generated consistently.

---

pkg/api/v1alpha1/types.go
hhiroshell: Same answer — under a normal Kubebuilder workflow, `controller-gen` would derive `DeepCopyObject`/`DeepCopyInto` from these type definitions automatically. Since we hand-wrote the types without a code-gen step, all three files are manual. A separate issue will track migrating to `controller-gen` to reduce maintenance burden and enable `+kubebuilder:validation` markers for admission-time enforcement.

---

pkg/controller/pod_controller_test.go
hhiroshell: Agreed — "as before" requires knowing the pre-PR implementation state, which is not fair to ask of new contributors. Will rename to something self-contained that describes what the test actually verifies:

```go
t.Run("annotation-based warmup: uses WarmupExecutor when no WarmupConfigName is set", func(t *testing.T) {
```

---

pkg/warmup/scenario_executor.go
hhiroshell: Correct — there is no upper bound check. The annotation-based warmup caps at `MaxTimeout = 5 minutes` (config.go:25), but `spec.Timeout` in WarmupConfig has no ceiling, so an arbitrary duration string (e.g. `"9999h"`) is accepted and silently clamped only by Go's `int64` nanosecond overflow (≈290 years).

Will add a consistent cap in `ExecuteScenario`:

```go
const maxScenarioTimeout = 5 * time.Minute

if overallTimeout > maxScenarioTimeout {
    overallTimeout = maxScenarioTimeout
}
```

Same cap applies to the per-step timeout, which should additionally be bounded by the remaining overall timeout rather than the fixed default. Will also add a `maximum` to `spec.timeout` and `steps[].timeout` in the CRD schema to catch obviously wrong values at admission time.

---

pkg/warmup/scenario_executor_test.go
hhiroshell: Good catch — `_ = step1Done` discards the value so we never assert it. The handler sets `step1Done = true` only after the request context is cancelled (via `case <-r.Context().Done()`), but since the test just discards it, we never confirm the cancellation actually happened.

A stronger test would assert both outcomes:

```go
result := e.ExecuteScenario(context.Background(), config, spec)
if !step2Done {
    t.Error("step 2 should have executed after step 1 timed out")
}
// Give the handler goroutine a moment to finish cleanup
time.Sleep(10 * time.Millisecond)
if !step1Done {
    t.Error("step 1 handler should have exited after step context was cancelled")
}
```

Or alternatively, check that `result.RequestsFailed >= 1` to confirm the timed-out step counted as a failure. Will improve the test.

---

docs/USAGE.md
hhiroshell: So, create an issue to address this.

docs/USAGE.md
hhiroshell: Also, create an issue to address this.

pkg/controller/pod_controller.go
hhiroshell: **Nil panic in WarmupConfig not-found fallback**: fix

Fixed in commit a8cfe84. The fallback path no longer calls `r.WarmupExecutor.Execute` — it now returns a `&warmup.Result{Success: false}` directly. This eliminates the nil dereference and also corrects the dispatch semantics (WarmupConfig not found → fail open directly, not silently fall back to annotation-based warmup).

---

pkg/warmup/http_sender.go
hhiroshell: **Unbounded io.ReadAll**: fix

Fixed in commit c387609. Added `io.LimitReader(resp.Body, 1<<20)` (1 MiB cap) before `io.ReadAll`. Responses larger than 1 MiB are silently truncated; this is acceptable because warmup responses used for extraction should be small.

---

pkg/warmup/session.go
hhiroshell: **Non-deterministic map iteration / template injection**: fix

Fixed in commit 00689b2. `Interpolate` now collects all keys, sorts them, and iterates in sorted order. This ensures deterministic substitution and prevents a substituted value containing `{{key}}` from being re-substituted by a later key.

---

pkg/warmup/scenario_executor.go
hhiroshell: **ScenarioExecutorIface naming**: fix

Fixed in commit fa0464f. The interface is now `ScenarioExecutor`; the concrete struct is `defaultScenarioExecutor`. All option function signatures, the compile-time assertion, and references in `mock.go` and `pod_controller.go` are updated.

---

pkg/warmup/scenario_executor.go
hhiroshell: **Inline URL construction duplicating BuildEndpointURL logic**: fix

Fixed in commits 92fdef9 and fa0464f. Added `Config.BuildEndpointURLFor(endpoint string)` in config.go (handles empty path → "/" and missing leading slash), then replaced the inline `fmt.Sprintf` in `executeStep` with `config.BuildEndpointURLFor(endpoint)`. The redundant prefix/slash check in `executeStep` was removed.

---

pkg/warmup/scenario_executor.go
hhiroshell: **HTTPSender allocated inside count loop**: fix

Fixed in commit fa0464f. `httpSender := &HTTPSender{client: e.httpClient, logger: e.logger}` is now constructed once before the `for i := 0; i < count; i++` loop, matching the GRPCSender pattern (which is already created once per step, outside the loop).

---

pkg/api/v1alpha1/deepcopy.go
hhiroshell: **deepcopy.go is hand-written instead of controller-gen generated**: new issue

Created issue #71 to track migrating to controller-gen: https://github.com/hhiroshell/kube-booster/issues/71

In the meantime, added mutation tests in 3511e6b that verify `DeepCopy` and `DeepCopyInto` fully isolate slices and maps — so any future field addition that breaks deep copy will fail the test rather than silently sharing state.

---

config/crd/warmupconfig.yaml
hhiroshell: **CRD missing maxLength and count maximum**: fix

Fixed in commit 1285443. Added:
- `maxLength: 32` on duration strings (timeout fields)
- `maxLength: 1024` on name, endpoint, grpcMethod, and extract value strings
- `maxLength: 65536` on body and grpcPayload (64 KiB)
- `maxLength: 16` on method (HTTP verb)
- `maxLength: 4096` on header values
- `maximum: 12000` on count
- `maxProperties: 50` on headers and extract maps

Also updated the spec.timeout description to document the 5-minute runtime cap.

---

pkg/warmup/scenario_executor.go
hhiroshell: **extractVariables as receiver method**: fix

Fixed in commit fa0464f. Converted to a package-level function `extractVariables(body []byte, extract map[string]string, session *SessionContext, logger logr.Logger, reqName string)`. The logger is now passed explicitly rather than captured through the receiver, making the function independently testable without constructing a full `defaultScenarioExecutor`.

---

pkg/controller/pod_controller.go
hhiroshell: **WarmupConfig lookup uses outer ctx instead of warmupCtx**: fix

Fixed in commit a8cfe84. Changed `r.Get(ctx, ...)` to `r.Get(warmupCtx, ...)`. The warmup context carries the configured warmup timeout, so the CR lookup is now bounded by the same deadline as the warmup execution itself.

---

docs/USAGE.md
hhiroshell: **JSONPath extraction limitation not clearly documented**: won't fix

No action taken: the limitation is already documented in USAGE.md just above this line — "**JSONPath extraction limitations:** Only simple dot-paths are supported (`$.key`, `$.a.b`). Array indexing and filter expressions are not supported." The docs are accurate.

---

docs/DEVELOPMENT.md
hhiroshell: **grafana-dashboard.json missing from directory tree**: fix

Fixed in commit 9a00f91. The file still exists at `config/samples/grafana-dashboard.json` on disk — it was accidentally dropped from the directory tree listing when the new sample files were added. Restored the entry.

---

docs/DEVELOPMENT.md
hhiroshell: **Custom headers doc line missing from http_sender.go section**: fix

Fixed in commit 9a00f91. Restored the line: `Custom headers: `User-Agent: kube-booster/1.0`, `X-Warmup-Request: true``. It was unintentionally dropped when the HTTPSender description was updated for the new response body capture feature.

---

docs/USAGE.md
hhiroshell: **Richer extraction format support**: new issue

Created issue #72 to track this: https://github.com/hhiroshell/kube-booster/issues/72

The issue covers adding an optional `format` selector to `extract` entries to support XML, regex, and gRPC field paths alongside the current JSONPath-only implementation.

---

docs/USAGE.md
hhiroshell: **"hit JIT compile thresholds" is too narrow**: fix

Fixed in commit 9a00f91. Updated to "warm up caches or trigger runtime optimization thresholds" which covers the JVM/GraalVM case but also caching, lazy init, and other use cases.

---

docs/USAGE.md
hhiroshell: **Auth pattern documentation for WarmupConfig credentials**: new issue

Created issue #73 to track this: https://github.com/hhiroshell/kube-booster/issues/73

The issue covers adding a "Security Considerations" subsection to the WarmupConfig docs covering short-lived tokens, service account patterns, dedicated unauthenticated warmup endpoints, and a future `secretRef` field.

---

docs/USAGE.md
hhiroshell: **WarmupConfig not-found fallback silently uses annotation-based warmup**: fix

Fixed in commit a8cfe84. The fallback no longer calls `r.WarmupExecutor.Execute`. When the WarmupConfig CR is not found, the controller now:
1. Logs the error
2. Emits a `WarmupFailed` warning event with the CR name
3. Returns `&warmup.Result{Success: false}` directly

The fail-open logic below then marks the pod READY with a `WarmupFailedOpen` reason — consistent with how other warmup failures are handled. USAGE.md is updated to document the new behavior.

---

pkg/api/v1alpha1/register.go
hhiroshell: **register.go is hand-written**: new issue

Same as deepcopy.go — covered by issue #71: https://github.com/hhiroshell/kube-booster/issues/71

Under a controller-gen workflow, the scheme registration boilerplate in register.go would also be derived from `+groupversion` markers rather than maintained manually.

---

pkg/api/v1alpha1/types.go
hhiroshell: **types.go is hand-written**: new issue

Same as deepcopy.go and register.go — covered by issue #71: https://github.com/hhiroshell/kube-booster/issues/71

Migrating to controller-gen would let `+kubebuilder:validation` markers in types.go drive both the CRD schema (replacing the hand-maintained YAML) and any generated code, eliminating the current three-way sync requirement between types.go, deepcopy.go, and warmupconfig.yaml.

---

pkg/controller/pod_controller_test.go
hhiroshell: **Test name "no WarmupConfigName uses WarmupExecutor as before" requires prior context**: fix

Fixed in commit a8cfe84. Renamed to `"annotation-based warmup: uses WarmupExecutor when no WarmupConfigName is set"` — fully self-describing without needing to know the pre-PR state.

Also renamed the adjacent test from `"WarmupConfig not found falls back to WarmupExecutor"` to `"WarmupConfig not found fails open without calling WarmupExecutor"` to match the updated dispatch semantics.

---

pkg/warmup/scenario_executor.go
hhiroshell: **No upper bound on scenario overallTimeout**: fix

Fixed in commit fa0464f. Added `maxScenarioTimeout = 5 * time.Minute` constant and a cap immediately after parsing:

```go
if overallTimeout > maxScenarioTimeout {
    overallTimeout = maxScenarioTimeout
}
```

Also added `maximum: 12000` on the CRD `count` field (commit 1285443) and `maxLength: 32` on duration string fields to reject obviously malformed inputs at admission time.

---

pkg/warmup/scenario_executor_test.go
hhiroshell: **_ = step1Done never asserts anything**: fix

Fixed in commit f3df271. After `ExecuteScenario` returns, the test now:

```go
if !step2Done {
    t.Error("step 2 should have executed after step 1 timed out")
}
time.Sleep(10 * time.Millisecond)
if !step1Done {
    t.Error("step 1 handler should have exited after step context was cancelled")
}
```

The 10ms sleep gives the handler goroutine time to observe the cancellation and set `step1Done = true`.

---

hhiroshell (commented):
LGTM

PR #74, opened 2026-07-02
feat: add Prometheus and Grafana monitoring sample manifests

hhiroshell:
## Summary

- Adds `config/samples/monitoring.yaml` — a single-file monitoring stack that users can apply with one command to get Prometheus and Grafana observability for kube-booster
- Prometheus is deployed directly (no Operator) with Kubernetes pod service discovery scoped to `app.kubernetes.io/name=kube-booster` on port 8080, plus the four alerting rules from `docs/OBSERVABILITY.md`
- Grafana is deployed with the existing `config/samples/grafana-dashboard.json` dashboard pre-provisioned via ConfigMap, and the Prometheus datasource auto-configured

## Usage

```bash
kubectl apply -f config/samples/monitoring.yaml

# Access Grafana (dashboard is pre-loaded)
kubectl port-forward -n monitoring svc/grafana 3000:3000
```

## Test plan

- [ ] `kubectl apply -f config/samples/monitoring.yaml` creates all 12 resources without errors
- [ ] Prometheus pod reaches Running state and the kube-booster scrape target appears under Status → Targets
- [ ] Grafana pod reaches Running state; the "kube-booster Warmup Metrics" dashboard is visible at `http://localhost:3000` after port-forwarding
- [ ] Alerting rules load cleanly under Prometheus Status → Rules

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config/samples/monitoring.yaml
+++ b/config/samples/monitoring.yaml
@@ -68,6 +68,17 @@ data:
 
   rules.yaml: |
     groups:
+      - name: kube-booster-recording
+        interval: 15s
+        rules:
+          - record: kube_booster_warmup_total:rate5m
+            expr: sum(rate(kube_booster_warmup_total[5m])) by (namespace, result)
+          - record: kube_booster_warmup_duration_seconds_bucket:rate5m
+            expr: sum(rate(kube_booster_warmup_duration_seconds_bucket[5m])) by (le, namespace)
+          - record: kube_booster_warmup_queue_wait_seconds_bucket:rate5m
+            expr: sum(rate(kube_booster_warmup_queue_wait_seconds_bucket[5m])) by (le, namespace)
+          - record: kube_booster_warmup_requests_total:rate5m
+            expr: sum(rate(kube_booster_warmup_requests_total[5m])) by (namespace)
       - name: kube-booster
         rules:
           - alert: KubeBoosterHighWarmupFailureRate
@@ -322,7 +333,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "sum(rate(kube_booster_warmup_total{result=\"success\", namespace=~\"$namespace\"}[5m])) / sum(rate(kube_booster_warmup_total{namespace=~\"$namespace\"}[5m]))",
+              "expr": "sum(kube_booster_warmup_total:rate5m{result=\"success\", namespace=~\"$namespace\"}) / sum(kube_booster_warmup_total:rate5m{namespace=~\"$namespace\"})",
               "legendFormat": "Success Rate",
               "refId": "A"
             }
@@ -442,7 +453,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "sum(rate(kube_booster_warmup_total{namespace=~\"$namespace\"}[5m])) by (result) * 60",
+              "expr": "sum(kube_booster_warmup_total:rate5m{namespace=~\"$namespace\"}) by (result) * 60",
               "legendFormat": "{{result}}",
               "refId": "A"
             }
@@ -532,7 +543,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "histogram_quantile(0.50, sum(rate(kube_booster_warmup_duration_seconds_bucket{namespace=~\"$namespace\"}[5m])) by (le))",
+              "expr": "histogram_quantile(0.50, sum(kube_booster_warmup_duration_seconds_bucket:rate5m{namespace=~\"$namespace\"}) by (le))",
               "legendFormat": "P50",
               "refId": "A"
             },
@@ -541,7 +552,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "histogram_quantile(0.95, sum(rate(kube_booster_warmup_duration_seconds_bucket{namespace=~\"$namespace\"}[5m])) by (le))",
+              "expr": "histogram_quantile(0.95, sum(kube_booster_warmup_duration_seconds_bucket:rate5m{namespace=~\"$namespace\"}) by (le))",
               "legendFormat": "P95",
               "refId": "B"
             },
@@ -550,7 +561,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "histogram_quantile(0.99, sum(rate(kube_booster_warmup_duration_seconds_bucket{namespace=~\"$namespace\"}[5m])) by (le))",
+              "expr": "histogram_quantile(0.99, sum(kube_booster_warmup_duration_seconds_bucket:rate5m{namespace=~\"$namespace\"}) by (le))",
               "legendFormat": "P99",
               "refId": "C"
             }
@@ -648,7 +659,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "histogram_quantile(0.50, sum(rate(kube_booster_warmup_queue_wait_seconds_bucket{namespace=~\"$namespace\"}[5m])) by (le))",
+              "expr": "histogram_quantile(0.50, sum(kube_booster_warmup_queue_wait_seconds_bucket:rate5m{namespace=~\"$namespace\"}) by (le))",
               "legendFormat": "P50",
               "refId": "A"
             },
@@ -657,7 +668,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "histogram_quantile(0.95, sum(rate(kube_booster_warmup_queue_wait_seconds_bucket{namespace=~\"$namespace\"}[5m])) by (le))",
+              "expr": "histogram_quantile(0.95, sum(kube_booster_warmup_queue_wait_seconds_bucket:rate5m{namespace=~\"$namespace\"}) by (le))",
               "legendFormat": "P95",
               "refId": "B"
             },
@@ -666,7 +677,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "histogram_quantile(0.99, sum(rate(kube_booster_warmup_queue_wait_seconds_bucket{namespace=~\"$namespace\"}[5m])) by (le))",
+              "expr": "histogram_quantile(0.99, sum(kube_booster_warmup_queue_wait_seconds_bucket:rate5m{namespace=~\"$namespace\"}) by (le))",
               "legendFormat": "P99",
               "refId": "C"
             }
@@ -911,7 +922,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "sum(rate(kube_booster_warmup_requests_total{namespace=~\"$namespace\"}[5m]))",
+              "expr": "sum(kube_booster_warmup_requests_total:rate5m{namespace=~\"$namespace\"})",
               "legendFormat": "Request Rate",
               "refId": "A"
             }
@@ -1000,7 +1011,7 @@ data:
                 "type": "prometheus",
                 "uid": "${datasource}"
               },
-              "expr": "sum(rate(kube_booster_warmup_total{namespace=~\"$namespace\"}[5m])) by (namespace, result) * 60",
+              "expr": "sum(kube_booster_warmup_total:rate5m{namespace=~\"$namespace\"}) by (namespace, result) * 60",
               "legendFormat": "{{namespace}} - {{result}}",
               "refId": "A"
             }
--- a/config/samples/monitoring.yaml
+++ b/config/samples/monitoring.yaml
@@ -161,6 +161,12 @@ spec:
           limits:
             cpu: 500m
             memory: 512Mi
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 65534
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["ALL"]
       volumes:
       - name: config
         configMap:
@@ -1110,6 +1116,12 @@ spec:
           limits:
             cpu: 200m
             memory: 256Mi
+        securityContext:
+          runAsNonRoot: true
+          runAsUser: 472
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: ["ALL"]
       volumes:
       - name: datasources
         configMap:
--- a/config/samples/monitoring.yaml
+++ b/config/samples/monitoring.yaml
@@ -7,32 +7,29 @@ metadata:
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: prometheus
+  name: kube-booster-prometheus
   namespace: monitoring
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: prometheus
+  name: kube-booster-prometheus
 rules:
 - apiGroups: [""]
   resources: ["pods", "nodes", "endpoints", "services"]
   verbs: ["get", "list", "watch"]
-- apiGroups: [""]
-  resources: ["configmaps"]
-  verbs: ["get"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: prometheus
+  name: kube-booster-prometheus
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: prometheus
+  name: kube-booster-prometheus
 subjects:
 - kind: ServiceAccount
-  name: prometheus
+  name: kube-booster-prometheus
   namespace: monitoring
 ---
 apiVersion: v1
@@ -133,7 +130,7 @@ spec:
       labels:
         app: prometheus
     spec:
-      serviceAccountName: prometheus
+      serviceAccountName: kube-booster-prometheus
       containers:
       - name: prometheus
         image: prom/prometheus:v3.4.1
--- a/config/samples/monitoring.yaml
+++ b/config/samples/monitoring.yaml
@@ -80,6 +80,7 @@ data:
                 /
                 sum(rate(kube_booster_warmup_total[5m]))
               ) > 0.1
+              and sum(rate(kube_booster_warmup_total[5m])) > 0
             for: 5m
             labels:
               severity: warning
--- a/config/samples/monitoring.yaml
+++ b/config/samples/monitoring.yaml
@@ -40,7 +40,7 @@ metadata:
 data:
   prometheus.yml: |
     global:
-      scrape_interval: 15s
+      scrape_interval: 15s      # increase to 30s or 60s for large clusters
       evaluation_interval: 15s
 
     rule_files:
@@ -180,7 +180,8 @@ spec:
         configMap:
           name: prometheus-config
       - name: storage
-        emptyDir: {}
+        emptyDir: {}  # ephemeral — for production use a PersistentVolumeClaim
+                      # and pass --storage.tsdb.retention.size to Prometheus
 ---
 apiVersion: v1
 kind: Service
@@ -224,7 +225,7 @@ data:
       - name: kube-booster
         type: file
         options:
-          path: /var/lib/grafana/dashboards
+          path: /var/lib/grafana/dashboards  # must match mountPath for grafana-dashboard-kube-booster in the Grafana Deployment
 ---
 apiVersion: v1
 kind: ConfigMap
@@ -1106,6 +1107,8 @@ spec:
         - containerPort: 3000
           name: web
         env:
+        # For development/demo only — disable anonymous access and configure
+        # proper authentication before deploying to shared or production clusters.
         - name: GF_AUTH_ANONYMOUS_ENABLED
           value: "true"
         - name: GF_AUTH_ANONYMOUS_ORG_ROLE
--- a/config/samples/monitoring.yaml
+++ b/config/samples/monitoring.yaml
@@ -1,0 +1,1135 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: monitoring
+---
+# Prometheus – ServiceAccount, RBAC, config, and deployment
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: prometheus
+  namespace: monitoring
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: prometheus
+rules:
+- apiGroups: [""]
+  resources: ["pods", "nodes", "endpoints", "services"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: prometheus
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: prometheus
+subjects:
+- kind: ServiceAccount
+  name: prometheus
+  namespace: monitoring
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus-config
+  namespace: monitoring
+data:
+  prometheus.yml: |
+    global:
+      scrape_interval: 15s
+      evaluation_interval: 15s
+
+    rule_files:
+      - /etc/prometheus/rules/rules.yaml
+
+    scrape_configs:
+      - job_name: 'kube-booster'
+        kubernetes_sd_configs:
+          - role: pod
+        relabel_configs:
+          # Keep only kube-booster pods
+          - source_labels: [__meta_kubernetes_pod_label_app_kubernetes_io_name]
+            regex: kube-booster
+            action: keep
+          # Keep only the metrics port (8080)
+          - source_labels: [__meta_kubernetes_pod_container_port_number]
+            regex: "8080"
+            action: keep
+          # Use pod name as instance label
+          - source_labels: [__meta_kubernetes_pod_name]
+            target_label: instance
+          # Preserve namespace label
+          - source_labels: [__meta_kubernetes_namespace]
+            target_label: kubernetes_namespace
+
+  rules.yaml: |
+    groups:
+      - name: kube-booster
+        rules:
+          - alert: KubeBoosterHighWarmupFailureRate
+            expr: |
+              (
+                sum(rate(kube_booster_warmup_total{result="failure"}[5m]))
+                /
+                sum(rate(kube_booster_warmup_total[5m]))
+              ) > 0.1
+            for: 5m
+            labels:
+              severity: warning
+            annotations:
+              summary: "High warmup failure rate"
+              description: "More than 10% of warmups are failing in the last 5 minutes"
+          - alert: KubeBoosterSlowWarmup
+            expr: |
+              histogram_quantile(0.95, sum(rate(kube_booster_warmup_duration_seconds_bucket[5m])) by (le))
+              > 60
+            for: 5m
+            labels:
+              severity: warning
+            annotations:
+              summary: "Slow warmup executions"
+              description: "P95 warmup duration is over 60 seconds"
+          - alert: KubeBoosterWarmupBacklog
+            expr: sum(kube_booster_warmup_active_pods) > 10
+            for: 5m
+            labels:
+              severity: warning
+            annotations:
+              summary: "Warmup backlog building up"
+              description: "More than 10 pods are actively executing warmup"
+          - alert: KubeBoosterHighSemaphoreWaitTime
+            expr: |
+              histogram_quantile(0.95, sum(rate(kube_booster_warmup_queue_wait_seconds_bucket[5m])) by (le))
+              > 10
+            for: 5m
+            labels:
+              severity: warning
+            annotations:
+              summary: "High warmup semaphore wait time"
+              description: "P95 wait time for the warmup semaphore exceeds 10 seconds — consider increasing --max-concurrent-warmups"
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: prometheus
+  namespace: monitoring
+  labels:
+    app: prometheus
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: prometheus
+  template:
+    metadata:
+      labels:
+        app: prometheus
+    spec:
+      serviceAccountName: prometheus
+      containers:
+      - name: prometheus
+        image: prom/prometheus:v3.4.1
+        args:
+        - --config.file=/etc/prometheus/prometheus.yml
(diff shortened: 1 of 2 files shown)


config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+        options:
+          path: /var/lib/grafana/dashboards
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: grafana-dashboard-kube-booster
+  namespace: monitoring
+data:
+  kube-booster.json: |
```
hhiroshell, line 236: Use Kustomize's configMapGenerator to keep this manifest from becoming too long.

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+  name: prometheus
+  namespace: monitoring
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus-config
+  namespace: monitoring
+data:
+  prometheus.yml: |
```
hhiroshell: alertmanagerとつなぐ設定を書いてコメントアウトしておいて。

hhiroshell (commented):
Test review

hhiroshell (commented):
Test

hhiroshell (commented):

This review was generated by analyzing PR #74 from 8 perspectives.

### Summary

| Perspective | Findings |
|-------------|----------|
| Logic & Correctness | 4 issues |
| Design & Maintainability | 6 issues |
| Security | 5 issues |
| Performance | 5 issues |
| Convention Compliance | 1 issue |
| Git History Context | 0 issues |
| PR History Context | 3 issues (all informational) |
| Documentation | 4 issues |
| **Total** | **28 issues** |

### Key Findings

#### :rotating_light: Errors (must fix)

1. **[Logic] Alert division-by-zero** (line 79): `KubeBoosterHighWarmupFailureRate` can produce `NaN` when no warmups have occurred — guard the expression with `and sum(rate(kube_booster_warmup_total[5m])) > 0`.

2. **[Design] Dashboard JSON duplicated** (line 220): The Grafana dashboard JSON embedded in the `grafana-dashboard-kube-booster` ConfigMap is a verbatim copy of the existing `config/samples/grafana-dashboard.json`. Future dashboard updates must be applied in two places and will silently drift.

3. **[Security/Logic/Design] Grafana anonymous Admin** (lines 1094–1096): `GF_AUTH_ANONYMOUS_ORG_ROLE=Admin` grants any cluster workload that can reach the Grafana ClusterIP full administrative control over Grafana.

#### :warning: Warnings

4. **[Logic] Histogram bucket mismatch** (line 89): `KubeBoosterSlowWarmup` fires at 60 s but `kube_booster_warmup_duration_seconds` uses `prometheus.DefBuckets` (largest bucket: 10 s) — `histogram_quantile` extrapolates beyond the last bucket, making the 60 s threshold unreliable.

5. **[Security] No `securityContext` on Prometheus container** (line 138): `prom/prometheus` defaults to root (UID 0) if not overridden.

6. **[Logic/Security/Performance] `emptyDir` for Prometheus TSDB** (line 168): Metrics are lost on pod restart (alert `for:` timers reset), storage is unbounded, and data is unencrypted at rest.

7. **[Design] Generic RBAC resource names** (line 14): `ClusterRole` and `ClusterRoleBinding` named `prometheus` (cluster-scoped) will overwrite existing resources in clusters that already run Prometheus.

8. **[Security] No `securityContext` on Grafana container** (line 1089): UID not explicitly enforced by the manifest.

9. **[Performance] Independent `histogram_quantile` queries** (line 531): Each P50/P95/P99 quantile per panel recalculates from raw bucket series on every 30 s refresh — recording rules would substantially reduce Prometheus query load.

#### :information_source: Documentation Gaps (files not in this diff)

The following docs should be updated as a follow-up (or in this PR):

- **`docs/OBSERVABILITY.md` line 237**: Add a "Quick Start" section for `monitoring.yaml` (one-command deploy), and clarify it embeds the same dashboard as `grafana-dashboard.json`.
- **`docs/DEVELOPMENT.md`**: Add `monitoring.yaml` to the project structure listing under `config/samples/`.
- **`README.md`**: Reference `monitoring.yaml` for quick observability setup.

> **Git history note**: All metric names in the PromQL queries are current and correct (including the rename from `kube_booster_pods_pending_warmup` → `kube_booster_warmup_active_pods` in commit `bacb0a6`). No historical regressions found.


---

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+---
+# Prometheus – ServiceAccount, RBAC, config, and deployment
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: prometheus
+  namespace: monitoring
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
hhiroshell, line 14: **[Design]** :warning: The `ClusterRole` and `ClusterRoleBinding` use the bare name `prometheus` with no project-specific prefix. These are cluster-scoped resources — applying this manifest to a cluster that already has a `prometheus` ClusterRole (e.g., from an existing Prometheus installation) will silently overwrite its rules. Consider naming them `kube-booster-prometheus`. The same issue applies to the `grafana` resources later in the file.

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+metadata:
+  name: prometheus
+rules:
+- apiGroups: [""]
+  resources: ["pods", "nodes", "endpoints", "services"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["configmaps"]
```
hhiroshell: **[Security]** :information_source: The Prometheus `ClusterRole` grants `get` on `configmaps` cluster-wide. If Prometheus only needs to scrape kube-booster pods via Kubernetes SD, this permission is broader than necessary and could expose ConfigMaps (including credentials stored as ConfigMaps) across all namespaces. Consider removing the `configmaps` entry if it is not strictly required for pod service discovery (CWE-272 – Least Privilege Violation).

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus-config
+  namespace: monitoring
+data:
+  prometheus.yml: |
+    global:
+      scrape_interval: 15s
```
hhiroshell: **[Performance]** :warning: `scrape_interval: 15s` is relatively aggressive. For clusters with many namespaces or high pod churn, this generates significant Prometheus storage and query pressure. Consider increasing to `30s` unless sub-minute precision is required for kube-booster metrics.

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+            target_label: kubernetes_namespace
+
+  rules.yaml: |
+    groups:
+      - name: kube-booster
+        rules:
+          - alert: KubeBoosterHighWarmupFailureRate
+            expr: |
+              (
+                sum(rate(kube_booster_warmup_total{result="failure"}[5m]))
```
hhiroshell, line 87: **[Logic]** :rotating_light: `KubeBoosterHighWarmupFailureRate` divides `sum(rate(kube_booster_warmup_total{result="failure"}[5m]))` by `sum(rate(kube_booster_warmup_total[5m]))`. When no warmup events have occurred (startup, quiet periods), the denominator is 0 or `NaN`, causing the expression to evaluate to `NaN > 0.1` → silently false. Guard with an `and` clause to avoid unexpected behavior:

```yaml
expr: |
  (
    sum(rate(kube_booster_warmup_total{result="failure"}[5m]))
    /
    sum(rate(kube_booster_warmup_total[5m]))
  ) > 0.1
  and sum(rate(kube_booster_warmup_total[5m])) > 0
```

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+                /
+                sum(rate(kube_booster_warmup_total[5m]))
+              ) > 0.1
+            for: 5m
+            labels:
+              severity: warning
+            annotations:
+              summary: "High warmup failure rate"
+              description: "More than 10% of warmups are failing in the last 5 minutes"
+          - alert: KubeBoosterSlowWarmup
```
hhiroshell, line 98: **[Logic]** :warning: `KubeBoosterSlowWarmup` fires when `histogram_quantile(0.95, ...)` on `kube_booster_warmup_duration_seconds_bucket` exceeds 60 s. However, `kube_booster_warmup_duration_seconds` is defined with `prometheus.DefBuckets` in `pkg/metrics/metrics.go`, whose largest boundary is **10 s**. `histogram_quantile` linearly extrapolates beyond the last bucket, making any P95 estimate for durations > 10 s highly inaccurate. The alert threshold of 60 s will be computed from an unreliable extrapolated value.

Either extend the histogram buckets in `pkg/metrics/metrics.go` to cover at least 60–120 s (similar to `kube_booster_warmup_queue_wait_seconds` which already extends to 300 s), or lower the alert threshold to within the defined bucket range (e.g., > 8s).

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+      app: prometheus
+  template:
+    metadata:
+      labels:
+        app: prometheus
+    spec:
+      serviceAccountName: prometheus
+      containers:
+      - name: prometheus
+        image: prom/prometheus:v3.4.1
```
hhiroshell, line 147: **[Security]** :warning: The Prometheus Deployment does not specify a `securityContext`. The `prom/prometheus` image runs as root (UID 0) by default. A container breach therefore yields root privileges, increasing the blast radius of any container-escape vulnerability. Add:

```yaml
securityContext:
  runAsNonRoot: true
  runAsUser: 65534
  readOnlyRootFilesystem: true
  allowPrivilegeEscalation: false
  capabilities:
    drop: [ALL]
```

(CWE-250, OWASP A05:2021 – Security Misconfiguration)

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+            memory: 256Mi
+          limits:
+            cpu: 500m
+            memory: 512Mi
+      volumes:
+      - name: config
+        configMap:
+          name: prometheus-config
+      - name: storage
+        emptyDir: {}
```
hhiroshell: **[Logic]** **[Security]** **[Performance]** :warning: Prometheus TSDB is backed by `emptyDir` with no size limit. Three separate concerns:

- **Logic**: All scraped metrics and TSDB state are lost on pod restart (OOMKill, eviction, rolling update). Alert `for: 5m` pending timers reset on restart, so active incidents can be silently dropped.
- **Security**: `emptyDir` bypasses PersistentVolume access controls and StorageClass encryption policies — metrics data is unencrypted at rest.
- **Performance**: Without `--storage.tsdb.retention.size`, metrics can grow unchecked and consume all node disk space.

For non-throwaway deployments, use a PersistentVolumeClaim and pass `--storage.tsdb.retention.size=<limit>`.

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+  name: grafana-dashboards-provider
+  namespace: monitoring
+data:
+  dashboards.yaml: |
+    apiVersion: 1
+    providers:
+      - name: kube-booster
+        type: file
+        options:
+          path: /var/lib/grafana/dashboards
```
hhiroshell: **[Design]** :information_source: The `grafana-dashboards-provider` ConfigMap hardcodes the dashboard path as `/var/lib/grafana/dashboards`, and the Grafana Deployment mounts `grafana-dashboard-kube-booster` at that same path. This implicit coupling is invisible at the ConfigMap level — if either path is changed without updating the other, dashboards silently disappear. A comment cross-referencing the two resources would make the contract explicit.

config/samples/monitoring.yaml
hhiroshell, line 236: **[Design]** :rotating_light: The Grafana dashboard JSON embedded here (lines 220–1069) is a verbatim duplicate of the existing `config/samples/grafana-dashboard.json`. This violates the single-source-of-truth principle: future dashboard updates must be applied in two places and will silently drift.

Consider generating the ConfigMap at deploy time instead of embedding:

```bash
kubectl create configmap grafana-dashboard-kube-booster \
  --from-file=kube-booster.json=config/samples/grafana-dashboard.json \
  -n monitoring --dry-run=client -o yaml
```

Alternatively, provide a setup script that creates the ConfigMap from the canonical JSON file and remove the embedded copy.

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+              "sort": "none"
+            }
+          },
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "${datasource}"
+              },
+              "expr": "sum(rate(kube_booster_warmup_total{result=\"success\", namespace=~\"$namespace\"}[5m])) / sum(rate(kube_booster_warmup_total{namespace=~\"$namespace\"}[5m]))",
```
hhiroshell: **[Performance]** :warning: The Warmup Success Rate panel divides two independent `sum(rate(...))` queries over the full dataset. Each query requires Prometheus to aggregate across all time series, and they run independently on every panel refresh. Consider a single `sum(...) by (result)` query with post-processing in Grafana, or define a recording rule to precompute the rates.

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+              "sort": "none"
+            }
+          },
+          "targets": [
+            {
+              "datasource": {
+                "type": "prometheus",
+                "uid": "${datasource}"
+              },
+              "expr": "histogram_quantile(0.50, sum(rate(kube_booster_warmup_duration_seconds_bucket{namespace=~\"$namespace\"}[5m])) by (le))",
```
hhiroshell: **[Performance]** :warning: The Warmup Duration panel contains three independent `histogram_quantile(0.50/0.95/0.99, sum(rate(kube_booster_warmup_duration_seconds_bucket[5m])) by (le))` targets — each recalculating the full bucket rate aggregation from raw series. With a 30 s auto-refresh and 9 panels, this generates repeated expensive computations. Consider a Prometheus recording rule:

```yaml
- record: job:kube_booster_warmup_duration_seconds_bucket:rate5m
  expr: sum(rate(kube_booster_warmup_duration_seconds_bucket[5m])) by (le)
```

then reference `histogram_quantile(0.95, job:kube_booster_warmup_duration_seconds_bucket:rate5m)` in the dashboard panels.

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+              "expr": "sum(rate(kube_booster_warmup_total{namespace=~\"$namespace\"}[5m])) by (namespace, result) * 60",
+              "legendFormat": "{{namespace}} - {{result}}",
+              "refId": "A"
+            }
+          ],
+          "title": "Warmup Rate by Namespace",
+          "type": "timeseries"
+        }
+      ],
+      "refresh": "30s",
```
hhiroshell, line 1024: **[Performance]** :information_source: The dashboard auto-refreshes every 30 s (`"refresh": "30s"`). With 9 panels each issuing separate Prometheus queries, this generates up to 270 query executions per hour per open browser tab. For large clusters with high-cardinality metrics, consider setting the default to `60s` or leaving it unset (`""`) to let users choose.

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+    matchLabels:
+      app: grafana
+  template:
+    metadata:
+      labels:
+        app: grafana
+    spec:
+      containers:
+      - name: grafana
+        image: grafana/grafana:12.0.1
```
hhiroshell, line 1105: **[Security]** :warning: The Grafana Deployment does not specify a `securityContext`. While recent `grafana/grafana` images default to UID 472, this is not enforced by the manifest. Without explicit constraints, a permissive cluster admission policy could allow privilege escalation. Add:

```yaml
securityContext:
  runAsNonRoot: true
  runAsUser: 472
  allowPrivilegeEscalation: false
  capabilities:
    drop: [ALL]
```

(CWE-250, OWASP A05:2021 – Security Misconfiguration)

config/samples/monitoring.yaml
```diff
@@ -0,0 +1,1135 @@
...
+        ports:
+        - containerPort: 3000
+          name: web
+        env:
+        - name: GF_AUTH_ANONYMOUS_ENABLED
+          value: "true"
+        - name: GF_AUTH_ANONYMOUS_ORG_ROLE
```
hhiroshell, line 1114: **[Security]** **[Logic]** **[Design]** :rotating_light: `GF_AUTH_ANONYMOUS_ENABLED=true` combined with `GF_AUTH_ANONYMOUS_ORG_ROLE=Admin` grants any unauthenticated entity that can reach the Grafana ClusterIP (any workload in the cluster, or any user via `kubectl port-forward`) full administrative privileges — including modifying dashboards, data sources, and executing arbitrary PromQL queries.

For development convenience this is understandable, but it should be made explicit:
- At minimum, change to `GF_AUTH_ANONYMOUS_ORG_ROLE: "Viewer"`
- For non-development environments, disable anonymous access entirely and configure proper authentication
- Add a comment like `# WARNING: admin anonymous access — for local dev only` so users know before applying to shared clusters

config/samples/monitoring.yaml
hhiroshell: Extend the bucket like below:
```
var WarmupBuckets = []float64{.5, 1, 2.5, 5, 10, 15, 30, 45, 60, 90, 120}
```

config/samples/monitoring.yaml
hhiroshell: There is no need to make it backed by a persistent volume. It's just an example manifest.

config/samples/monitoring.yaml
hhiroshell: The purpose of this manifest is simply to provide an example and demonstration of kube-booster monitoring. Therefore, we prioritize ease of use over strict access control.

config/samples/monitoring.yaml
hhiroshell: Great suggestion. `configMapGenerator` would reference `grafana-dashboard.json` directly and solve both the length issue and the duplication problem raised in the other thread at this line. A `kustomization.yaml` in `config/samples/` could look like:

```yaml
configMapGenerator:
- name: grafana-dashboard-kube-booster
  namespace: monitoring
  files:
    - kube-booster.json=grafana-dashboard.json
generatorOptions:
  disableNameSuffixHash: true
```

That way `grafana-dashboard.json` stays the single source of truth. Agree this is the right structural approach — will restructure accordingly.

---

config/samples/monitoring.yaml
hhiroshell: Agreed — collision risk is real since `ClusterRole` and `ClusterRoleBinding` are cluster-scoped. Renaming to `kube-booster-prometheus` and `kube-booster-grafana` (and updating all references) is a straightforward no-functional-impact fix. Will address this.

---

config/samples/monitoring.yaml
hhiroshell: Fair point. The `configmaps` permission was included following the scrape config example in `docs/OBSERVABILITY.md`, but Kubernetes pod service discovery does not actually require it. Removing it brings the ClusterRole in line with least-privilege. Will drop that entry.

---

config/samples/monitoring.yaml
hhiroshell: For a demo manifest, 15s is useful to see metrics appear quickly after deploying kube-booster. The 30s suggestion is valid for production. Will add a comment to the config making this trade-off explicit so users know to increase it before deploying at scale.

---

config/samples/monitoring.yaml
hhiroshell: Agreed — this is a genuine logic issue. The correct fix is to guard the expression:

```yaml
expr: |
  (
    sum(rate(kube_booster_warmup_total{result="failure"}[5m]))
    /
    sum(rate(kube_booster_warmup_total[5m]))
  ) > 0.1
  and sum(rate(kube_booster_warmup_total[5m])) > 0
```

Will update.

---

config/samples/monitoring.yaml
hhiroshell: The proposed buckets are well-chosen — they cover the typical warmup range and extend to 120 s, well above the `KubeBoosterSlowWarmup` threshold of 60 s, so `histogram_quantile` will give accurate results rather than extrapolating.

The change in `pkg/metrics/metrics.go` is straightforward — replace `prometheus.DefBuckets` at line 32 with:

```go
Buckets: []float64{.5, 1, 2.5, 5, 10, 15, 30, 45, 60, 90, 120},
```

Note that this is a breaking change to the metric's histogram schema (existing bucket series will not line up with new ones after a rolling restart), but since this is still pre-stable it's the right time to fix it. Will make this change in `pkg/metrics/metrics.go` and update the alert threshold comment in `monitoring.yaml` accordingly.

---

config/samples/monitoring.yaml
hhiroshell: Valid — and since this manifest is likely to be copy-pasted, modeling best practices here avoids passing insecure defaults downstream. Will add:

```yaml
securityContext:
  runAsNonRoot: true
  runAsUser: 65534  # nobody
  readOnlyRootFilesystem: true
  allowPrivilegeEscalation: false
  capabilities:
    drop: [ALL]
```

The Prometheus data directory (`/prometheus`) and rules directory need a writable `emptyDir` volume mount, which is already present.

---

config/samples/monitoring.yaml
hhiroshell: Agreed — PVC is unnecessary overhead for an example. Will add a short comment near the `emptyDir` entry so users who copy-paste to production know what they're getting into:

```yaml
- name: storage
  emptyDir: {}  # ephemeral — for production use a PersistentVolumeClaim
                # and pass --storage.tsdb.retention.size to Prometheus
```

---

config/samples/monitoring.yaml
hhiroshell: Agreed — a brief cross-reference comment on both the provider ConfigMap and the Deployment volume mount makes the contract explicit. Will add something like:

```yaml
# path must match the mountPath for grafana-dashboard-kube-booster in the Grafana Deployment
path: /var/lib/grafana/dashboards
```

Note: this coupling also disappears if we adopt the Kustomize `configMapGenerator` approach suggested in the other thread — the path would still exist but the configmap name mismatch risk goes away.

---

config/samples/monitoring.yaml
hhiroshell: This is directly resolved by adopting the Kustomize `configMapGenerator` approach proposed in the other thread. With `configMapGenerator` referencing `grafana-dashboard.json`, the embedded copy disappears entirely and `grafana-dashboard.json` becomes the single source of truth — no manual sync needed on future dashboard updates.

---

config/samples/monitoring.yaml
hhiroshell: For a sample dashboard, the impact is negligible — this pattern is standard in Grafana and Prometheus handles it fine at typical kube-booster deployment scale. Recording rules are a production optimization to add when the cluster grows large enough that dashboard refresh becomes noticeable in Prometheus CPU/memory. Will note this in a comment near the panel rather than pre-optimizing in the sample.

---

config/samples/monitoring.yaml
hhiroshell: Same rationale as the success rate panel — acceptable for a sample. The more pressing issue here is the bucket range, which is being addressed in the line 89 thread. Once `WarmupDurationSeconds` uses the extended buckets, the P50/P95/P99 queries here will give accurate results. Recording rules remain a nice-to-have for large clusters.

---

config/samples/monitoring.yaml
hhiroshell: 30 s is a reasonable default for a demo — it lets users see metrics update promptly. Grafana users can always change the interval interactively via the dashboard toolbar. Leaving it as-is.

---

config/samples/monitoring.yaml
hhiroshell: Consistent with the Prometheus securityContext decision at line 138 — will add it here as well to model best practices. Grafana defaults to UID 472; will enforce it explicitly:

```yaml
securityContext:
  runAsNonRoot: true
  runAsUser: 472
  allowPrivilegeEscalation: false
  capabilities:
    drop: [ALL]
```

---

config/samples/monitoring.yaml
hhiroshell: Understood — ease of use is the right call for a demo manifest. Will add a comment near these env vars to document the intent and prevent silent misuse:

```yaml
env:
# For development/demo only — disable anonymous access and configure
# proper authentication before deploying to shared or production clusters.
- name: GF_AUTH_ANONYMOUS_ENABLED
  value: "true"
```

---

config/samples/monitoring.yaml
hhiroshell: But if it's an easy fix, please go ahead and include it in this PR. I believe providing a well-configured query example is another purpose of this example manifest.

config/samples/monitoring.yaml
hhiroshell: But if it's an easy fix, please go ahead and include it in this PR. I believe providing a well-configured query example is another purpose of this example manifest.

config/samples/monitoring.yaml
hhiroshell: **Dashboard JSON duplication / Kustomize restructuring**: new issue

Created issue #75 to track this: https://github.com/hhiroshell/kube-booster/issues/75

The issue proposes replacing the embedded JSON with a `kustomization.yaml` using `configMapGenerator`, referencing `grafana-dashboard.json` as the single source of truth.

---

config/samples/monitoring.yaml
hhiroshell: **Generic RBAC resource names**: fix

Fixed in commit 5807d80. Renamed `ClusterRole`, `ClusterRoleBinding`, and `ServiceAccount` from `prometheus` to `kube-booster-prometheus` and updated all internal references (Deployment `serviceAccountName`, `ClusterRoleBinding` `roleRef`).

---

config/samples/monitoring.yaml
hhiroshell: **Overly-broad `configmaps` permission**: fix

Fixed in commit 5807d80. Removed the `configmaps` rule from the `ClusterRole` — it is not required for Kubernetes pod service discovery.

---

config/samples/monitoring.yaml
hhiroshell: **`scrape_interval: 15s`**: won't fix

No functional change — 15s is appropriate for a demo manifest. Added an inline comment in commit 6629e55 noting users should increase to 30s–60s for large clusters.

---

config/samples/monitoring.yaml
hhiroshell: **Alert division by zero**: fix

Fixed in commit fef279d. Added `and sum(rate(kube_booster_warmup_total[5m])) > 0` guard to the `KubeBoosterHighWarmupFailureRate` expression.

---

config/samples/monitoring.yaml
hhiroshell: **Histogram bucket mismatch**: fix

Fixed in commit c91dea2. Changed `pkg/metrics/metrics.go` line 32 from `prometheus.DefBuckets` to `[]float64{.5, 1, 2.5, 5, 10, 15, 30, 45, 60, 90, 120}` — matching the proposed buckets. The `KubeBoosterSlowWarmup` 60s threshold is now well within the bucket range, so `histogram_quantile` will give accurate results.

---

config/samples/monitoring.yaml
hhiroshell: **Missing Prometheus `securityContext`**: fix

Fixed in commit de9249e. Added `runAsNonRoot: true`, `runAsUser: 65534`, `allowPrivilegeEscalation: false`, and `capabilities: drop: [ALL]` to the Prometheus container.

---

config/samples/monitoring.yaml
hhiroshell: **Ephemeral `emptyDir` storage**: won't fix

No functional change per owner's decision — PVC is unnecessary for a demo. Added a comment in commit 6629e55 marking the `emptyDir` as intentionally ephemeral and pointing users to PVC + `--storage.tsdb.retention.size` for production.

---

config/samples/monitoring.yaml
hhiroshell: **Implicit path coupling**: fix

Fixed in commit 6629e55. Added an inline comment on the `path` field cross-referencing the `grafana-dashboard-kube-booster` ConfigMap mount in the Grafana Deployment.

---

config/samples/monitoring.yaml
hhiroshell: **Embedded dashboard JSON duplication**: new issue

Tracked in issue #75: https://github.com/hhiroshell/kube-booster/issues/75

Addressed structurally via the Kustomize `configMapGenerator` approach described there.

---

config/samples/monitoring.yaml
hhiroshell: **Expensive independent rate queries**: fix

Fixed in commit 26764ce. Added a `kube_booster_warmup_total:rate5m` recording rule and updated the Warmup Success Rate and Warmup Rate panels to reference it, eliminating redundant per-refresh aggregation.

---

config/samples/monitoring.yaml
hhiroshell: **Repeated histogram bucket rate aggregation**: fix

Fixed in commit 26764ce. Added `kube_booster_warmup_duration_seconds_bucket:rate5m` and `kube_booster_warmup_queue_wait_seconds_bucket:rate5m` recording rules. All six P50/P95/P99 panel expressions across the Warmup Duration and Semaphore Queue Wait panels now reference the precomputed recording rules.

---

config/samples/monitoring.yaml
hhiroshell: **30s dashboard auto-refresh**: won't fix

No change — 30s is appropriate for a demo. With the recording rules now in place (commit 26764ce), the per-refresh Prometheus load is substantially reduced, making this less of a concern even at larger scale.

---

config/samples/monitoring.yaml
hhiroshell: **Missing Grafana `securityContext`**: fix

Fixed in commit de9249e. Added `runAsNonRoot: true`, `runAsUser: 472`, `allowPrivilegeEscalation: false`, and `capabilities: drop: [ALL]` to the Grafana container.

---

config/samples/monitoring.yaml
hhiroshell: **Anonymous Admin access**: won't fix

No functional change per owner's decision — ease of use is prioritized for a demo manifest. Added a dev/demo-only warning comment in commit 6629e55 above the `GF_AUTH_ANONYMOUS_*` env vars.

---

hhiroshell (commented):
LGTM. Let's do the tests on an actual cluster after we resolve #75.

